PR #1, opened 2016-10-18
Bug fixes for python3.5

danbgray:
- As of python3.2, xmlrpclib becomes xmlrpc.client,
- Removed .pyc files
- In some instances was running into a KeyIndex 'name' not found.  Added exception handling.


tonybaloney:
thanks @danbgray hope this has been helpful for you. Good catch on the module import, will also add future to requirements

PR #23, opened 2017-04-17
revert "publisher: ensure used space name is uppercase"

jdknight:
This reverts commit 4f85cd030374b8e1bafd2bf49c0e49a598008de6.

Observed on Confluence Cloud that requests are case-sensitive to the
space of a name. A user must match the configured space name to the case
of the Confluence space name.

Signed-off-by: James Knight <james.d.knight@live.com>


tonybaloney:
@jdknight 👍

PR #28, opened 2017-05-05
fix python3 < 3.6 incompatibility in rest.py

tmalch:
use rsp.text instead of rsp.content for json.loads.

Prior to python3.6 json.loads only allows type str or unicode (python2).
rsp.content is of type bytes in python3 so json.loads fails (python3.* < 3.6).
rsp.text is of type str in python3, unicode in python2, it should work.


tonybaloney:
thanks for your contribution @tmalch

PR #36, opened 2017-06-28
experimental support for confluence page hierarchy

JPTeasdale:
This change adds the `confluence_experimental_page_hierarchy` option. When True, the builder will scan the `master_doc` for any toctree and recursively set up parent docs. On publish, the parent of each doc in confluence will be set to it's parent in the toctree. 

Adds functionality from https://github.com/tonybaloney/sphinxcontrib-confluencebuilder/issues/35

### DIFF
--- a/sphinxcontrib/confluencebuilder/builder.py
+++ b/sphinxcontrib/confluencebuilder/builder.py
@@ -14,6 +14,7 @@ from .publisher import ConfluencePublisher
 from .writer import ConfluenceWriter
 from docutils.io import StringOutput
 from docutils import nodes
+from sphinx import addnodes
 from sphinx.builders import Builder
 from sphinx.util.osutil import ensuredir, SEP
 from os import path
@@ -112,6 +113,10 @@ class ConfluenceBuilder(Builder):
         else:
             self.space_name = None
 
+        if self.config.master_doc and self.config.confluence_experimental_page_hierarchy:
+            self.register_parents(self.config.master_doc)
+
+
     def get_outdated_docs(self):
         """
         Return an iterable of input files that are outdated.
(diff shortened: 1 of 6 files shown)


sphinxcontrib/confluencebuilder/publisher.py
```diff
@@ -283,6 +283,7 @@ def storePage(self, page_name, raw_data, parent_id=None):
                         newPage['ancestors'] = [{'id': parent_id}]
 
                     rsp = self.rest_client.post('content', newPage)
+                    uploaded_page_id = rsp['id']
```
JPTeasdale: Had to add this in, twas a bug in the function where it wouldn't return the uploaded_id in this scenario.

test/test_builder.py
```diff
@@ -152,11 +157,38 @@ def test_table(self):
...
+    def setUpClass(self):
+        basedir = os.path.dirname(os.path.realpath(__file__))
+        srcdir = os.path.join(basedir, 'testproj')
+        self.expected = os.path.join(srcdir, 'expected')
+        builddir = os.path.join(srcdir, 'build')
+        self.outdir = os.path.join(builddir, 'out')
+        doctreedir = os.path.join(builddir, 'doctree')
+
+        self.config = create_default_test_config()
+        self.config['master_doc'] = "toctree"
```
JPTeasdale: this causes the following warning to display. We don't want to include contents.rst in any toctree because that would muck up other tests, so I don't know what to do about this. Test passes though. 

`/Users/johnteasdale/Uber/sphinxcontrib-confluencebuilder/test/testproj/contents.rst: WARNING: document isn't included in any toctree`

sphinxcontrib/confluencebuilder/builder.py
```diff
@@ -151,7 +152,18 @@ def get_relative_uri(self, from_, to, typ=None):
         return relative_uri(self.get_target_uri(from_),
                             self.get_target_uri(to, typ))
 
+    def register_parents(self, filename):
+        tree = self.env.get_doctree(filename)
+        for toctreenode in tree.traverse(addnodes.toctree):
+            for includefile in toctreenode['includefiles']:
```
tonybaloney: is `includefiles` always in the dictionary?

JPTeasdale: Sphinx core makes that assumption: https://github.com/sphinx-doc/sphinx/blob/master/sphinx/util/nodes.py

Check `inline_all_toctrees`

sphinxcontrib/confluencebuilder/publisher.py
jdknight: :+1:

jdknight:
I'll closing this PR due to the new/combined PR #38.

PR #37, opened 2017-06-28
Inline all documents after a max_depth

JPTeasdale:
In this change, if you set `confluence_experimental_max_depth`, the builder will render all toctree imports beyond the posted number inline in the page. For example `confluence_experimental_max_depth=0` will have similar behavior to the 'singlehtml' builder in sphinx proper. 

No tests in this one yet. I'll update this after we merge the the page_hierarchy one if you want to go in that direction.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ doc/_build/
 # testing
 .tox/
 .test_publish_key
+
+env/*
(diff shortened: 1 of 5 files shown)


jdknight:
I'll closing this PR due to the new/combined PR #38.

Issue #41, opened 2017-08-11
Curly bracket inside command causes 500 error.


sabinlinden:
This line in a .rst:
`* Host is located at ``/var/lib/store/hosts/{hostname}``.
`

Caused this error:
```
sphinxcontrib.confluencebuilder error:
---
An unsupported Confluence API call has been made.

REQ: POST
RSP: 500
URL: https://company.atlassian.net/wiki/rest/api/
API: contentbody/convert/storage
---
```

Removing the curly brackets around hostname made the error go away.

jdknight:
Related to #39.

tonybaloney:
Fixed in #43  
![image](https://user-images.githubusercontent.com/1532417/29504281-ed242262-8681-11e7-92e1-100870a4f27e.png)

tonybaloney:
@sabinlinden please can you validate this is fixed. 

`pip install -e git+https://github.com/tonybaloney/sphinxcontrib-confluencebuilder.git#egg=sphinxcontrib-confluencebuilder`

sabinlinden:
Unfortunately the pip install command errors out in my current environment and I have yet to figure out the reason why. I'll let you know if I figure this out.

spyoungtech:
@sabinlinden try upgrading setuptools. 

```
pip install --upgrade setuptools
```

sabinlinden:
Sorry, that didn't work and I've been strapped for time so I still haven't resolved the issue with my environment.

jdknight:
This issue should now be resolved in db5b916e57898eb86244b4c33a801f5da5c3b984 (available on the master; see specifically ca84f547796c091d80a131bc2edb8b5bfd435441 AND b6bb29230db22ccc45a5852c75f88ef4cfbea422). To take advantage of this change, one can pull from `master`. This change should be made available in an upcoming v0.7 release (#29).

Issue #40, opened 2017-08-03
Unsupported API call has been made


cblegare:
When uploading docs

### Versions

- Confluence 6.0.6 (Trial license)
- sphinxcontrib-confluencebuilder version `0.7.0.dev0` (installed with `pip install git+https://github.com/tonybaloney/sphinxcontrib-confluencebuilder`)

### Bad HTML

I think the issue is there

```html
<ri:page ri:content-title=''build', 'dist', '* ... '.coverage', 'htmlcov''></ri:page>
```

Where there are single quotes enclosed within singles quotes

In this case, here is the `.conf` content

```text
**class setup.Clean(dist, **kw)**

   Custom clean command to tidy up the project.

   ``default_patterns = ['build', 'dist', '* ... '.coverage', 'htmlcov']``
```

which is generated from the following RestructuredText

```rst
.. autoclass:: setup.Clean

   .. autoattribute:: setup.Clean.default_patterns
```

### Traceback

```text
Traceback (most recent call last):
  File "/home/me/stuff/venv/lib/python3.6/site-packages/sphinx/cmdline.py", line 306, in main
    app.build(opts.force_all, filenames)
  File "/home/me/stuff/venv/lib/python3.6/site-packages/sphinx/application.py", line 339, in build
    self.builder.build_update()
  File "/home/me/stuff/venv/lib/python3.6/site-packages/sphinx/builders/__init__.py", line 331, in build_update
    'out of date' % len(to_build))
  File "/home/me/stuff/venv/lib/python3.6/site-packages/sphinx/builders/__init__.py", line 397, in build
    self.write(docnames, list(updated_docnames), method)
  File "/home/me/stuff/venv/lib/python3.6/site-packages/sphinx/builders/__init__.py", line 434, in write
    self._write_serial(sorted(docnames))
  File "/home/me/stuff/venv/lib/python3.6/site-packages/sphinx/builders/__init__.py", line 443, in _write_serial
    self.write_doc(docname, doctree)
  File "/home/me/stuff/venv/lib/python3.6/site-packages/sphinxcontrib/confluencebuilder/builder.py", line 218, in write_doc
    self.publish_doc(docname, self.writer.output)
  File "/home/me/stuff/venv/lib/python3.6/site-packages/sphinxcontrib/confluencebuilder/builder.py", line 226, in publish_doc
    uploaded_id = self.publisher.storePage(title, output, self.parent_id)
[...]

Issue #39, opened 2017-07-07
Unescaped markup breaks publishing


spyoungtech:
I ran into an issue attempting to publish an rst document which, in relevant part, contains a list that has url's that contain placeholder text (IE `<placeholder>`) which resulted in a
`ConfluenceBadApiError` exception.

Checking the response returned by Confluence:
```py
{"statusCode":400,"data":{"authorized":false,"valid":true,"errors":[],"successful":false},"message":"Error parsing xhtml: Unexpected close tag </li>; expected </project-name>.\n at [row,col {unknown-source}]: [139,67]"}
```

Sample of the RST content:

```rst
Deployment
----------

We will discuss this topic in depth,
but here are some links for your further reference
when the class is complete:

* It should be possible to export the contents of ``_build/html``
  to any file-system-based web service and serve it as static content.

* You can package the documentation in a ZIP file
  and upload it using the “edit” page for your Python package,
  and it will appear at the URL:

  http://pythonhosted.org/<project-name>

  Detailed instructions for this procedure live at:

  http://pythonhosted.org/<project-name>

* The powerful and popular Read the Docs service
  lets you configure your GitHub repository
  so that every time you push a new version of your software,
  the documentation gets automatically rebuilt
  and made available at:

  https://readthedocs.org/projects/<project-name>/

  Read the Docs also supports custom host names
  if you want your documentation to appear beneath your own
  project sub-domain.
  More information is available at:
```

jdknight:
@spyoungtech, thanks for the sample case. Definitely an issue handling the greater-than/less-than characters. Both REST and XML-RPC will fail.

I tried making a quick fix escaping some characters, however `contentbody/convert/storage` (and `convertWikiToStorageFormat`) appear to unescape the data again before a storage request is made. So, there may be a need to two-passes to escape some characters (from just a quick look at this). When this issue can be looked at, it would be good to check for additional characters which could cause this issue as well.

tonybaloney:
thanks for filing the issue @spyoungtech I'll hopefully get around to reviewing this next week

tonybaloney:
This is a tricky one, I'm reading the spec and `<`, `>` are reserved characters so it's not clear whether your document is even valid RST

http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#inline-markup-recognition-rules

I'm reluctant to just put a blanket `replace` method in for those characters since they can be used for other signs and it'll just create new bugs.

[...]

PR #43, opened 2017-08-21
Escape bug fix

tonybaloney:
Fixes for #39  #40  #41


jdknight:
:+1:

tonybaloney:
The bugs are actually 3 different issues. 
#41 is where code blocks need to be escaped. That can easily be fixed.
#40 I don't understand yet
#39 is when you have a link to an invalid URL.

PR #46, opened 2017-09-29
publisher: allow publish without notification

jdknight:
In Confluence, when a page is updated, watchers will be notified of any page changes unless the edit is flagged as minor. The following provides the ability for a publisher to flag any page updates as minor edits. By setting the configuration option `confluence_disable_notifications` to `True`, any existing page edits will be flagged as a "minor edit".

For XML-RPC publishing, page updates are now handled by the "updatePage" method (over the "storePage" method). This is to allow the publisher to provide update options along side any page content edits.


jdknight:
While this PR can be merged it, it is recommended to consider #49 first.

Issue #52, opened 2017-10-04
 unexpected "garbage characters" with quotation mark | â€™


jdknight:
In some cases, published Confluence pages may display undesired "garbage characters" inside documents. This is repeatable with a document such as:
````
:orphan:

oddity
======

Let's see this issue.
````
Where the rendered page would have the line:

> Letâ€™s see this issue.

This was a result of our received REST JSON data was interpreted from the Confluence instance. The response headers do not explicitly indicate that the received content-type is `application/json; charset=utf-8`; instead, just `application/json`. To address this issue, a fix will be made to interpret all received JSON data from Confluence to be UTF-8. This should be fine in most cases since [RFC-4627](https://www.ietf.org/rfc/rfc4627.txt) dictates that `application/json`shall be UTF-based (and by default UTF-8) and Atlassian documents that [Confluence should always be UTF-8](https://confluence.atlassian.com/doc/configuring-character-encoding-174010.html).

jdknight:
I've added a commit fa206b18982ad7765fd23e57b47718c6795df5ac ("rest: interpret received data as utf-8"), pending to be a pull request. This should address issues with encoding. Since this is more of an informative issue, I'm going to preemptively mark it as closed. If anyone has addition comments or other concerns, feel free to make a comment.

jdknight:
err -- re-opening until fa206b18982ad7765fd23e57b47718c6795df5ac is merged into `master`.

tonybaloney:
@jdknight #53 is merged so I assume this issue is now resolved?

jdknight:
@tonybaloney Yup!

This issue should now be resolved in db5b916e57898eb86244b4c33a801f5da5c3b984 (available on the master). To take advantage of this change, one can pull from `master`. This change should be made available in an upcoming v0.7 release (#29).

Issue #29, opened 2017-05-06
prepare a v0.7 release


jdknight:
_Note: No expected release date planned at this time._

The following issue captures the overall intent to release v0.7 of the `sphinxcontrib-confluencebuilder` module. This includes issues and more found under [0.7 milestone](https://github.com/tonybaloney/sphinxcontrib-confluencebuilder/milestone/2).

Before this release is made, the following tests are desired to be performed:

Owner/Contributor Todos:
- [x] Ensure all issues and pull requests in [0.7 milestone](https://github.com/tonybaloney/sphinxcontrib-confluencebuilder/milestone/2) have been addressed.
- [x] (optional) Ensure [pluto-silly](https://github.com/tonybaloney/sphinxcontrib-confluencebuilder/commits/pluto-silly) branch is merged in (whenever desired).
- [x] (optional) Merge [brave-backpack](https://github.com/tonybaloney/sphinxcontrib-confluencebuilder/commits/brave-backpack) branch.
- [x] (optional) Merge [feature/readme-demo](https://github.com/tonybaloney/sphinxcontrib-confluencebuilder/tree/feature/readme-demo) branch.

Owner Todos:
- [x] Update `CHANGES.rst`, replacing `master` with release version and date.
- [x] Bumpversion to `release` (with tag).
- [x] Bumpversion to `minor`.
- [x] Push release cycle commits and tag to `origin/master`.
- [x] Update Read the Docs space.

Owner/Contributor Todos:
- [x] From release tag, build and publish to PyPI (using `twine`).
- [x] Close this issue.

jdknight:
@tonybaloney, any interest in doing a v0.7 release? I'm hoping to take advantage of some of the changes/fixed introduced recently.

tonybaloney:
yeah, totally, there are some lint and build errors though in #31 @jdknight

jdknight:
@tonybaloney, thanks for recent reviews/merges. I've just re-tested a series of checks on Confluence Cloud and Confluence Server, and I believe it's a good time for a 0.7 release.

So, if you're willing and have the time, mind performing the release steps on the `master` branch? While you can also update PyPI as well, I'm also willing to performing the publishing (if you're busy).

tonybaloney:
@jdknight all done and published

jdknight:
@tonybaloney, awesome! Thanks!

PR #53, opened 2017-10-04
fixes and cleanup for pre-0.7 work - "butter risky"

jdknight:
The following contains a series of improvements for the Confluence builder extension. Most notables includes:

- Corrects slew of markup/publishing errors observed in #39, #40 and #41 (see 255043422d25f8e437baf3182eafa6464a1ab288 and b6bb29230db22ccc45a5852c75f88ef4cfbea422).
- Corrects unexpected "garbage" encoding data in #52 (see fa206b18982ad7765fd23e57b47718c6795df5ac).
- Have title references emphasized (instead of strong'ed).

Perform validation set testing on Confluence 6.0.6 server (REST and XML-RPC) and Confluence Cloud  (REST and XML-RPC) with expected/improved results.


sphinxcontrib/confluencebuilder/common.py
```diff
@@ -75,3 +76,13 @@ def verbose(*args, **kwargs):
     def warn(*args, **kwargs):
         if ConfluenceLogger.logger:
             ConfluenceLogger.logger.warning(*args, **kwargs)
+
+    @staticmethod
+    def trace(title, data):
+        try:
+            with io.open('trace.log', 'a', encoding='utf-8') as file:
+                file.write(u'[%s]\n' % title)
```
tonybaloney: this could cause issues in py2.7

```
u'%s' % unicode('👉')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
UnicodeDecodeError: 'ascii' codec can't decode byte 0xf0 in position 0: ordinal not in range(128)
```

tonybaloney: that is, assuming title is a byte string. I would encode it first

sphinxcontrib/confluencebuilder/translator/wiki.py
```diff
@@ -26,10 +26,10 @@
 
 # Confluence encoding for special chars
 SPECIAL_VALUE_REPLACEMENTS = {
-    ('{', '&lcub;'),
-    ('}', '&rcub;'),
-    ('<', '&lt;'),
-    ('>', '&gt;')
+    ('{', '&#123;'),
```
tonybaloney: why was this changed? I remember having issues with code points.

sphinxcontrib/confluencebuilder/common.py
jdknight: Ah, I don't think I fully tested tracing in 2.7. I'll take a look and cleanup.

sphinxcontrib/confluencebuilder/translator/wiki.py
jdknight: Interesting. Before the change (from commit b6bb29230db22ccc45a5852c75f88ef4cfbea422), I got an error when using the curly brackets:
```
MSG: Error parsing xhtml: Undeclared general entity "lcub"
 at [row,col {unknown-source}]: [1,390]
```
I'll go back and try to re-produce the issue. I also wonder if any of this relates to the use of the `HTMLParser` (proposed for removal in 255043422d25f8e437baf3182eafa6464a1ab288).

sphinxcontrib/confluencebuilder/common.py
jdknight: For this trace method, the title value is an internal-only provided (i.e. not a document title). I wouldn't mind just leaving the method as-is and just assume a provided title is a compatible byte string for the running interpreter.

sphinxcontrib/confluencebuilder/translator/wiki.py
jdknight: Do you have any hints on your observed issues using decimal values? After trying to toggle the change, I always see an error generated when using the '\&lcub;' (and respective) values:

```
An unsupported Confluence API call has been made.

REQ: PUT
RSP: 400
URL: https://jdknight.atlassian.net/wiki/rest/api/
API: content
MSG: Error parsing xhtml: Undeclared general entity "lcub"
 at [row,col {unknown-source}]: [3,30]
```

Always observing errors in 2.7, 3.4-3.6 on both Confluence Server and Confluence Cloud when using '\&lcub;'. No issues with the decimal variant. Thoughts?

tonybaloney:
@jdknight ready for release

jdknight:
@tonybaloney Thanks! I'll do some re-testing to re-sync myself, take a look at the recent issues to see if there's anything critical and if everything good I can ping you to wrap up the owner tasks in #29 (if you thin that's a good plan).

Issue #35, opened 2017-06-26
Uploader flattens page hierarchy


JPTeasdale:
If you have a page hierarchy like so:
```
master_doc.rst
other_doc.rst
other_doc_includes/other_doc_part1.rst
other_doc_includes/other_doc_part2.rst
```

The uploader does not preserve the page hierarchy as expected and instead creates four documents at the root of the space.

JPTeasdale:
Hmm I guess there's no real way to make this extensible since you don't know if there is going to be an other_doc_2.rst that also includes other_doc_includes/other_doc_part1.rst

jdknight:
The only way I can see for one to _properly_ take on some level of page hierarchy support is to rely on Sphinx's [toctree](http://www.sphinx-doc.org/en/stable/markup/toctree.html) markup (please, correct me if I'm wrong). I do not believe this extension should be defining an elaborate hierarchy configuration (i.e. mapping per-page structure) or custom Sphinx markups in order to support some level of page hierarchy (while I don't think this was proposed, I'd rather explicitly rule it out now).

Taking a look at `toctree`, a capability could be added to that from a `master_doc` page and a defined `toctree` (and nested `toctree` markups), one could provide some level of page hierarchy support. For example, using [Kernel documentation](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation) as an example, there could be a mapping such as:

- Welcome to The Linux Kernel's documentation [ref](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/index.rst)
  - The Linux kernel user's and administrator's guide
    - Bisecting a bug
    - Kernel Support for miscellaneous (your favourite) Binary Formats v1.1
    - Linux Braille Console
    - Linux allocated devices (4.x+ version)
    - ...
  - Development tools for the kernel [ref](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/dev-tools/index.rst)
    - Coccinelle
    - The Kernel Address Sanitizer (KASAN) [ref](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/dev-tools/kasan.rst)
    - Using gcov with the Linux kernel
    - ...
  - ...

Some other points to consider:
- Support for page hierarchy should only be enabled if:
  - A configuration option indicates to enable support;
[...]

PR #58, opened 2017-12-05
slew of extension improvements - "forgotten autumn"

jdknight:
The following contains a series of improvements for the Confluence builder extension. Most notables includes:

- Support for page hierarchy (based off of comment/changes noted in #35).
- Fix come publishing corner/error handling cases.
- Refactor a series of classes to help separate/cleanup/document some internal logic.


sphinxcontrib/confluencebuilder/translator/wiki.py
```diff
@@ -165,7 +180,10 @@ def depart_rubric(self, node):
         self.end_state()
 
     def visit_compound(self, node):
-        pass
+        if self.apply_hierarchy_children_macro:
+            self.add_text('{children:depth=%s}' % self.tocdepth)
+            self.add_text(self.nl)
+            raise nodes.SkipNode
```
tonybaloney: What does the skip node statement do?

sphinxcontrib/confluencebuilder/state.py
```diff
@@ -0,0 +1,169 @@
...
+    def title(docname):
+        """
+        return the title value for a provided docname
+
+        See `registerTitle` for more information.
+        """
+        return ConfluenceState.doc2title.get(docname)
+
+    @staticmethod
+    def titleConflictCheck():
```
tonybaloney: 👍

sphinxcontrib/confluencebuilder/builder.py
```diff
@@ -177,10 +163,14 @@ def prepare_writing(self, docnames):
 
             doctitle = first_section[idx].astext()
             if not doctitle:
+                if self.publish:
+                    ConfluenceLogger.warn("document will not be published "
```
tonybaloney: Should we just generate a title? I mean I'd be concerned people would use this in CI/CD and never see that warning

jdknight: It's a possibility. I would lean towards having a more blunt auto-naming schema if this is desired. For example, `notitle-<docname>` or `autogen-<docname>` to prevent any collisions with randomly generated document names (or just a docname) with other pages that possibly shouldn't be overridden.

Maybe it would be ideal to have a configuration option which could disable such a capability as well (i.e. by default, automatically generate a title if none is set unless explicitly configured to not do so).

If this is desired, it'll probably be good to start a new issue to capture this new capability outside this PR. 🤔

sphinxcontrib/confluencebuilder/translator/wiki.py
jdknight: The purpose of the `nodes.SkipNode` is to drop any sub-processing of the compound type (i.e. the toc-tree wrapper). So, if `apply_hierarchy_children_macro` is configure, the child macro will be rendered in the document and no longer attempt to process the internals of the toctree. For the default case, the toctree type is entered and renders its internal elements using other `visit_*` calls (based on the internals of Sphinx). Since the method has a definition, there's no need to define a `pass`.

sphinxcontrib/confluencebuilder/builder.py
tonybaloney: Agreed. It can be a separate feature.

tonybaloney:
@jdknight do you want additional features for 0.8? Or can this be prepared as a release

jdknight:
@tonybaloney, if you want to push out a 0.8 release now, that's fine -- I have nothing else immediately available to contribute. Anything that I may have tagged in a 0.8 milestone can be pushed to 0.9.

Issue #62, opened 2017-12-07
Combination of backticks and tildes creates error.


sabinlinden:
One of my rst files was generating this error -

```
sphinxcontrib.confluencebuilder error:
---
An unsupported Confluence API call has been made.

REQ: POST
RSP: 400
URL: https://lindenlab.atlassian.net/wiki/rest/api/
API: content
MSG: Error parsing xhtml: Unexpected close tag </code>; expected </sub>.
 at [row,col {unknown-source}]: [200,121]
---
```

When I dug in to see what it was, after a lot of trial and error I discovered the error was only triggered with a very specific combination of `` and ~ characters. The line in question contained the following string:
```
``~/.devscripts`` & ``~/.bashrc``
```

It turns out if you remove any one of the backticks or any one of the tildes, it causes the error to go away. It's this very specific combination creating the error. The ampersand doesn't seem to have any effect, either, but I wanted to include the exact string causing the bug just in case.

jdknight:
This is an interesting bug. First look at it, it actually seems to be a Confluence parsing error. Using the markup (to reiterate):

```
``~/.devscripts`` & ``~/.bashrc``
```
While get converted by this extension to:
```
{{~/.devscripts}} & {{~/.bashrc}}
```
And with the current internal reliance to convert Wiki format to storage format, Confluence converts this section to (a detection of a code block and a subscript block):
```
<code><sub>/.devscripts</code> &amp; <code></sub>/.bashrc</code>
```
The improper HTML results in the 400 error. A user workaround I see is using the following:
```
``&#126;/.devscripts`` & ``&#126;/.bashrc``
```
Which gives a working conversion that can be published:
```
<code>&#126;/.devscripts</code> &amp; <code>&#126;/.bashrc</code>.
```
An internal workaround could be to detect the code block in the source, and if a code block's first character is a tilde, perform a conversion to the special HTML character. While it's not graceful, it might be worth it since having code blocks leading with tilde are (I would suspect be) common. Ideally, whenever an internal storage-based publisher is supported (if ever), this wouldn't be an issue.

jdknight:
Hmm... some more investigation reveals even more problems. Replacing the tilde character with other special characters (e.x. `*`) results in the same issue. A crude way of looking at it seems like Confluence will perform any text formatting styles on a first-pass and then perform a monospace conversion on another pass.

[...]

PR #71, opened 2018-01-02
Added escaping for dashes

iamjackg:
Hey everyone.

I added escaping for dashes because the Confluence markup interpreter gets confused by some edge cases like this one, and it prevents uploads:

```
Call this tool with the {{--foo}} and {{--bar}} parameters.
```

The dashes act as striketrough tags in Confluence wiki markup, whereas escaped dashes are always displayed as normal dashes, so this shouldn't break anything else.

### DIFF
--- a/sphinxcontrib/confluencebuilder/translator/wiki.py
+++ b/sphinxcontrib/confluencebuilder/translator/wiki.py
@@ -29,7 +29,8 @@ SPECIAL_VALUE_REPLACEMENTS = {
     ('{', '&#123;'),
     ('}', '&#125;'),
     ('<', '&#60;'),
-    ('>', '&#62;')
+    ('>', '&#62;'),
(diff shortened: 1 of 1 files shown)


jdknight:
@iamjackg, thanks for the PR. This is one of those times where multiple people have been fixing the same thing.  😓

The root of this issue has been captured in #62. I've also made changes in #72 (specifically, 0f0c784f99234c509f9bf6f98a7be5132d568f49) which should also cover the change you've introduced here. I'm going to recommend we close this PR (sorry) for the newer PR which covers the additional markup types.

iamjackg:
Hah! Well, that's obviously no problem at all. I'm glad you were on top of it already!

Thanks for looking into this. I might be back Soon™ to try and add some more formatting to the the wiki translator.

Thanks again for your work on this project!

PR #73, opened 2018-01-05
rest: output post/put json on rest api err

jdknight:
In the event that a REST API POST/PUT call fails due to a Confluence
API error state, dump the pushed JSON content to the exception message.

This is to assist in a series of development and user-reported scenarios
where content pushes return a basic exception with only a limited amount
of information to debug. Along with the content data being dumped, the
JSON block of data will also include other helpful information such as
the target key/title, page identifier and more.

Signed-off-by: James Knight <james.d.knight@live.com>


tonybaloney:
recommend that it check the verbosity of the sphinx command first. The log data could be written to a CI/CD server or service, where someone could view the logs.
Considering TravisCI as an example is public, we wouldn't want it to contain potentially private data.

PR #82, opened 2018-01-23
unit test rework and fixes

jdknight:
_(sorry, this got bigger than expected)_

The following introduces an newer way to run unit tests for this extension. The primary reason for the change is to limit the amount of clutter inside the `test` folder. Unit tests are now added to the primary test suite using [unittest's discover](https://docs.python.org/2/library/unittest.html#unittest.TestLoader.discover) capability. Individual unit tests can have their own folders which can hold a series of document/expected sets of files.

Along with this new unit test running, the previous `test_builder` and `test_publish` classes have been ported over (and also improved in some cases).

During the re-examination of these tests and corrections based off of Sphinx documentation, the test build always failed due to found errors inside the code. Thus, a series of fixes have been applied to the builder, writer and more.


jdknight:
This PR may need a conflict correction when #84 is merged in (recommending #84 first). Specifically, due to the introduction of 0d0a1432fb33caed22448bf61500a0088ec5fb57 and e88ef73cec5446d7f59789fc256e7e578a6501f7.

tonybaloney:
Thanks @jdknight  no merge conflicts

Issue #76, opened 2018-01-17
Incorrect Confluence API base URL and bug handling ConfluenceBadAPI Error


layoaster:
I was just trying to generate and publish my documentation when I got the following error:

```
Traceback (most recent call last):
  File "/home/layo/projects/venv/onbi-data/lib/python3.6/site-packages/sphinx/cmdline.py", line 305, in main
    opts.warningiserror, opts.tags, opts.verbosity, opts.jobs)
  File "/home/layo/projects/venv/onbi-data/lib/python3.6/site-packages/sphinx/application.py", line 234, in __init__
    self._init_builder()
  File "/home/layo/projects/venv/onbi-data/lib/python3.6/site-packages/sphinx/application.py", line 311, in _init_builder
    self.builder.init()
  File "/home/layo/projects/venv/onbi-data/lib/python3.6/site-packages/sphinxcontrib/confluencebuilder/builder.py", line 98, in init
    self.publisher.connect()
  File "/home/layo/projects/venv/onbi-data/lib/python3.6/site-packages/sphinxcontrib/confluencebuilder/publisher.py", line 67, in connect
    'limit': 1
  File "/home/layo/projects/venv/onbi-data/lib/python3.6/site-packages/sphinxcontrib/confluencebuilder/rest.py", line 52, in get
    raise ConfluenceBadApiError(self._format_error(rsp, key))
  File "/home/layo/projects/venv/onbi-data/lib/python3.6/site-packages/sphinxcontrib/confluencebuilder/rest.py", line 138, in _format_error
    err += "MSG: " + rsp.json()['message']
  File "/home/layo/projects/venv/onbi-data/lib/python3.6/site-packages/requests/models.py", line 892, in json
    return complexjson.loads(self.text, **kwargs)
  File "/usr/lib/python3.6/json/__init__.py", line 354, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python3.6/json/decoder.py", line 339, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python3.6/json/decoder.py", line 357, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```
[...]

Issue #69, opened 2017-12-18
Possible to use with Confluence cloud?


petebachant:
My team and I are using Confluence cloud with two-factor authentication, and would like to push some Sphinx docs. Has this use case ever been tested?

jdknight:
@petebachant, it's not supported at this time. Atlassian's documentation [indicates](https://confluence.atlassian.com/cloud/secure-your-account-with-two-step-verification-939505063.html):

> **API access tokens with two-step verification**
If you enable two-step verification on an account that is used by scripts or services to access Atlassian Cloud REST APIs, then that account won't be able to use a password for basic authentication against a REST API. We recommend that you use an API token instead, although an organization admin could exclude the relevant account from two-step verification. Read more about API tokens.

To use the REST API (as stated above), an API token will need to be used instead. Unfortunately, when I go to my Confluence Cloud instance, it states that "[API tokens are] not available just yet. We're working on it, so please try again in the future". When API tokens into either a more working beta state or completely supported, we'll make sure this capability is added.

jdknight:
@petebachant, got some good news. The Atlassian cloud instance I have access now has the ability to create API tokens. After creating a token and testing with the most recent release of this extension, I'm happy to say that API tokens already work. By using your account's Email as the username (`confluence_server_user`) and API token value as the password (`confluence_server_pass`), you should be able to authenticate when using the REST API (note: API token support is not available over XML-RPC).

That all being said, I'll try to update our documentation to explicitly indicate this (and probably even promote to use API tokens over using an account's password).

Closing this issue (and thanks for bringing this up).

petebachant:
Just getting back to this now. Thanks a lot for your help!

PR #91, opened 2018-02-01
Update configuration.rst

nvtkaszpir:
Fix incorrect variable name in server url section.
Give example for Atlassian Confluence Cloud (/wiki suffix).
Use capitalized Confluence Cloud.


nvtkaszpir:
It should help in such non fully documented examples like #69, and #76

PR #99, opened 2018-03-21
Two-way SSL support

loganasherjones:
This branch will implement client-side certificate support, so that users running a client-authenticated confluence server can utilize this package.

Here is what is left to be done:

- [x] Make pylint work
- [x] Add the ability to specify a certificate authority
- [x] Allow client certificates to be specified
- [x] Add support for encrypted client-side certificates
- [x] Documentation
- [x] Figure out if I can even test the XML-RPC interface with my local confluence, if so, make it work with XML-RPC interface also


loganasherjones:
Okay, I believe this is now ready to go. Short of any changes you all would like to see. I've tested this with the following:

Test Case | Rest Interface | XML-RPC Interface | Notes
----------- | ---------------- | ------------- | ------
No SSL | :heavy_check_mark:  | :heavy_check_mark: | All SSL info is just ignored
1 way SSL | :heavy_check_mark: | :heavy_check_mark: | Valid CA provided
1 way SSL - disable CA checking | :heavy_check_mark: | :heavy_check_mark: | Ignores CA correctly
1 way SSL - Client certs specified | :heavy_check_mark: | :heavy_check_mark: | No error, just ignores client certs
2 way SSL - Encrypted Client certs | :heavy_check_mark: | :heavy_check_mark: | Hooray
2 way SSL - Encrypted + ignore CA | :heavy_check_mark: | :heavy_check_mark: | Also works, though you do get a lot of warnings from requests
2 way SSL - Encrypted Client certs broken apart | :heavy_check_mark: | :heavy_check_mark: | Checked that `('/path/to/cert.pem', '/path/to/key.pem')` works
2 way SSL - Encrypted Bad CA + ignore | :heavy_check_mark: | :heavy_check_mark: | Fails as expected
2 way SSL - All the same things + unencrypted | :heavy_check_mark: | :heavy_check_mark: | Hooray
2 way SSL - Invalid certificate | :heavy_check_mark: | :heavy_check_mark: | Fails with ssl verification error.
2 way SSL - Bad Password | :heavy_check_mark: | :heavy_check_mark: | No stacktrace, just an error message.

sphinxcontrib/confluencebuilder/config.py
```diff
@@ -123,4 +123,36 @@ def validate(c, log=True):
 properly set with the publisher's Confluence username.
 """)
 
+            if c.confluence_ca_cert:
+                if (not os.path.isfile(c.confluence_ca_cert) and
```
tonybaloney: you don't need parenthesis on this line

sphinxcontrib/confluencebuilder/config.py
```diff
@@ -123,4 +123,36 @@ def validate(c, log=True):
...
+                    if log:
+                        ConfluenceLogger.error(
+"""missing certificate authority
+
+The option 'confluence_ca_cert' has been provided to find a certificate 
+authority file or path from a relative location. Ensure the value is set to a 
+proper file path.
+""")
+
+            if c.confluence_client_cert:
```
tonybaloney: good idea, I like this as an API +1

sphinxcontrib/confluencebuilder/config.py
```diff
@@ -123,4 +123,36 @@ def validate(c, log=True):
...
+                if isinstance(c.confluence_client_cert, tuple):
+                    cert_files = c.confluence_client_cert
+                else:
+                    cert_files = (c.confluence_client_cert,)
+
+                for cert_file in cert_files:
+                    if not os.path.isfile(cert_file):
+                        errState = True
+                        if log:
+                            ConfluenceLogger.error(
```
tonybaloney: Why not raise in this condition instead of having a special error condition and logic flow? If the error is non-recoverable then you should raise the error

sphinxcontrib/confluencebuilder/publisher.py
```diff
@@ -464,6 +471,18 @@ def __init__(self, server_url, proxy=None,
         self.https = (self.scheme == 'https')
         self.proxy = None
         self.timeout = timeout
+        if isinstance(client_cert, tuple) and len(client_cert) >= 2:
```
tonybaloney: this conditional is confusing. Recommend checking for len of client_cert once and once only, refactor this to validate that client_cert is tuple and break into a conditional statement or using a ternary operator.

sphinxcontrib/confluencebuilder/publisher.py
```diff
@@ -464,6 +471,18 @@ def __init__(self, server_url, proxy=None,
...
+            if len(client_cert) < 2:
+                self._certfile = client_cert[0]
+                self._keyfile = None
+            else:
+                self._certfile = client_cert[0]
```
tonybaloney: I would use native Python unpacking for this, eg. self._certfile, self._keyfile = client_cert

sphinxcontrib/confluencebuilder/publisher.py
```diff
@@ -510,10 +529,7 @@ def make_connection(self, host):
         # build connection
         if self.https:
             try:
-                context = None
-                if self.disable_ssl_validation:
-                    import ssl
-                    context = ssl._create_unverified_context()
+                context = self._setup_ssl_context()
```
tonybaloney: there is still a condition where user would want a non-verified SSL connection surely?

sphinxcontrib/confluencebuilder/rest.py
```diff
@@ -17,32 +19,91 @@
...
 
+class SSLAdapter(HTTPAdapter):
+    def __init__(self, cert, password=None, disable_validation=False,
+                 *args, **kwargs):
+        if isinstance(cert, tuple) and len(cert) >= 2:
+            if len(cert) < 2:
+                self._certfile = cert[0]
+                self._keyfile = None
+            else:
+                self._certfile = cert[0]
```
tonybaloney: Use native unpacking

tonybaloney:
@loganasherjones thanks for the contribution! I've made some recommendations about code fixes and logic conditions. This needs unit tests before we can merge it, look at the existing test suite and leverage the Tox runtime to see how things are working now..

sphinxcontrib/confluencebuilder/config.py
loganasherjones: I didn’t raise an error for consistencies sake. None of the above error conditions actually raise an error, they just set the `errorState` to `True` which in turn raises an error. I assumed this was what Sphinx wanted.

I’d be happy to change it, but I think I should also change all the other error conditions in this file as well.

sphinxcontrib/confluencebuilder/publisher.py
loganasherjones: I did think about this. Do you have a recommended place to format the client cert? 

My initial thought was during config validation. But I didn’t think it fit well there from an immutability standpoint.

sphinxcontrib/confluencebuilder/publisher.py
loganasherjones: This is still supported. If they specify to disable ssl support, we can still return a valid ssl context with `check_hostname` set to false and the `verify_mode` set to `ssl.CERT_NONE`.

sphinxcontrib/confluencebuilder/config.py
loganasherjones: Sorry, I'm used to the google style-guide. Would you prefer to see a backslash?

jdknight: Can't we just use `os.path.exists` instead? (I could be wrong)

sphinxcontrib/confluencebuilder/config.py
jdknight: I think what was submitted is correct (with respect to setting the error flag and logging the error). This is to allow multiple configuration items to be verified (and output at once) before a configuration exception (`ConfluenceConfigurationError`) is thrown (in the parent class).

sphinxcontrib/confluencebuilder/publisher.py
```diff
@@ -23,13 +23,15 @@
 from .exceptions import ConfluencePermissionError
 from .exceptions import ConfluenceProxyPermissionError
 from .exceptions import ConfluenceRemoteApiDisabledError
+from .exceptions import ConfluenceCertificateError
```
jdknight: (optional) I'd prefer if list imports were sorted in a lexicographical fashion.

sphinxcontrib/confluencebuilder/rest.py
```diff
@@ -8,7 +8,9 @@
 """
 
 import json
+import ssl
```
jdknight: (optional) I'd prefer if list imports were sorted in a lexicographical fashion.

sphinxcontrib/confluencebuilder/rest.py
```diff
@@ -17,32 +19,82 @@
 from .exceptions import ConfluenceProxyPermissionError
 from .exceptions import ConfluenceSeraphAuthenticationFailedUrlError
 from .exceptions import ConfluenceTimeoutError
+from .exceptions import ConfluenceSSLError
```
jdknight: (optional) I'd prefer if list imports were sorted in a lexicographical fashion.

sphinxcontrib/confluencebuilder/rest.py
```diff
@@ -17,32 +19,82 @@
...
+from .exceptions import ConfluenceCertificateError
 from .std.confluence import API_REST_BIND_PATH
 
+class SSLAdapter(HTTPAdapter):
```
jdknight: (optional) While I know PEP 8 (and maybe Google's Python guidelines) promote "CapWords" style, I'd prefer the convention `SslAdapter`. If you want to leave it as is or if @tonybaloney prefers the "CapWords" style, that's fine by me.

sphinxcontrib/confluencebuilder/rest.py
```diff
@@ -17,32 +19,82 @@
...
+        super(SSLAdapter, self).__init__(*args, **kwargs)
+
+    def init_poolmanager(self, *args, **kwargs):
+        context = ssl.create_default_context(ssl.Purpose.SERVER_AUTH)
+        try:
+            context.load_cert_chain(certfile=self._certfile,
+                                    keyfile=self._keyfile,
+                                    password=self._password)
+        except ssl.SSLError as ex:
+            ConfluenceCertificateError(ex)
```
jdknight: Isn't this missing `raise`?

loganasherjones:
Okay, I've updated with tests for the config portion. 

I still have an outstanding question on the parenthesis on line 127 of config. I'm happy to change it to match whatever style you want, if you want a backslash or can recommend a better way, I can change it.

I'm happy to answer more questions in regards to disabling the SSL validation during XML RPC publishing if you have questions.

As for tests for the rest/rpc publishing portion. I added the _ability_ to test these things in the `test_publisher.py`, however it looks like this script is meant to be used more for integration testing? So, I ran the tests against my local confluence, in No-SSL/1-way SSL/2-way SSL and they all worked as expected. I believe testing these changes on the jdknight confluence page requires configuration changes to the server, but in theory, it could be done. I looked for unit tests for the rest client or publisher, but didn't see any. I'm not sure if you wanted me to create these or if just adding support in the integration tests was enough.

Hopefully this is better!

sphinxcontrib/confluencebuilder/rest.py
```diff
@@ -17,32 +19,82 @@
...
+        if self._disable_validation:
+            context.check_hostname = False
+
+        kwargs['ssl_context'] = context
+        return super(SSLAdapter, self).init_poolmanager(*args, **kwargs)
+
+
```
jdknight: (trivial) No need for two lines.

test/test_publisher.py
```diff
@@ -45,12 +45,15 @@ def setUpClass(self):
...
-        self._conf('confluence_server_url',     'CB_URL', DEFAULT_TEST_URL)
-        self._conf('confluence_server_user',    'CB_USR', DEFAULT_TEST_USER)
-        self._conf('confluence_space_name',     'CB_SPC', DEFAULT_TEST_SPACE)
+        self._conf('confluence_parent_page',      'CB_PAR', DEFAULT_TEST_PARENT)
+        self._conf('confluence_publish_prefix',   'CB_PFX')
+        self._conf('confluence_server_pass',      'CB_PWD')
+        self._conf('confluence_server_url',       'CB_URL', DEFAULT_TEST_URL)
+        self._conf('confluence_server_user',      'CB_USR', DEFAULT_TEST_USER)
+        self._conf('confluence_space_name',       'CB_SPC', DEFAULT_TEST_SPACE)
+        self._conf('confluence_ca_cert',          'CB_CA')
```
jdknight: (note) Heads up that the `test_publisher.py` class is about to be re-written. You can leave these changes in if you'd like, but just note these changes may or may not exist in the future (if it matters to you).

test/unit-tests/common/test_config.py
```diff
@@ -65,3 +65,74 @@ def test_publish_config(self):
...
+        builder.config.confluence_space_name = "some_space_name"
+        with self.assertRaises(ConfluenceConfigurationError):
+            builder.init(suppress_conf_check=True)
+
+        self.assertEqual(len(builder.config.confluence_client_cert), 2)
+
+        builder.config.confluence_publish = False
+        builder.config.confluence_client_cert = None
+        builder.config.confluence_server_url = None
+        builder.config.confluence_space_name = None
```
jdknight: Missing trailing line?

sphinxcontrib/confluencebuilder/config.py
loganasherjones: Yay! Now it fits on one line :)

sphinxcontrib/confluencebuilder/rest.py
loganasherjones: Thanks! Nice catch.

loganasherjones:
Okie dokie. I think feedback from both @tonybaloney and @jdknight have been implemented. Let me know if there is anything else!

jdknight:
@loganasherjones, much appreciated. Everything looks good for me (unless I'm overlooking something 😅). I'll let @tonybaloney drive it from here, unless he states otherwise.

tonybaloney:
Yep, it's a 👍 from me. 

Great work. Merging

loganasherjones:
Thanks for all the help! Looking forward to the next release :)

PR #101, opened 2018-03-23
Fixed bug for 1-way SSL RPC calls.

loganasherjones:
This resolves an issue such that if the RPC interface is being
used and no client_certificate was provided, an error would be
raised. This prevents this issue.

Signed-off-by: Logan Jones <loganasherjones@gmail.com>


tonybaloney:
I recommend first create a test that demonstrates the issue (a failing test), commit that, and then commit the fix and see the test pass.

loganasherjones:
@jdknight this is ready for review when you are 😄

jdknight:
@loganasherjones, thanks for the fix; much appreciated.

loganasherjones:
No problem! Sorry it was broken in the first place.

PR #113, opened 2018-04-27
Warnings are supposed to be colored but on Windows they are sometimes

papadeltasierra:
not and are displayed like  '?[31;01mThis is a warning...'.
Colorama is not intercepting the control strings and converting them
into Windows color commands.

Colorama has to be initialized and the resulting sys.stderr mock
passed to Sphinx() to ensure correct colorization.

Signed-off-by: Paul D.Smith <paul_d_smith@hotmail.com>

### DIFF
--- a/test/sphinxcontrib_confluencebuilder_util.py
+++ b/test/sphinxcontrib_confluencebuilder_util.py
@@ -7,6 +7,7 @@
     :license: BSD, see LICENSE.txt for details.
 """
 
+from sphinx.util.console import nocolor, color_terminal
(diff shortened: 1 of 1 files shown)


test/sphinxcontrib_confluencebuilder_util.py
```diff
@@ -7,6 +7,7 @@
     :license: BSD, see LICENSE.txt for details.
 """
 
+from sphinx.util.console import nocolor, color_terminal
```
jdknight: (trivial; optional) import not in lexicography order

PR #114, opened 2018-05-11
Improvements to ToDo support.

papadeltasierra:
This fixes what appear to be some minor omissions/bugs in the ToDo support.  I am a little concerned about the Doctitle fix which seems to be needed but feels like someone should have spotted this earlier or there is something I'm missing.

### DIFF
--- a/sphinxcontrib/confluencebuilder/translator/wiki.py
+++ b/sphinxcontrib/confluencebuilder/translator/wiki.py
@@ -847,10 +847,6 @@ class ConfluenceWikiTranslator(ConfluenceTranslator):
             docname = posixpath.normpath(
                 self.docparent + path.splitext(node['refuri'])[0])
             doctitle = ConfluenceState.title(docname)
-            if not doctitle:
-                ConfluenceLogger.warn("unable to build link to document due to "
-                    "missing title (in %s): %s" % (self.docname, docname))
-                raise nodes.SkipNode
 
             if '#' in node['refuri']:
                 anchor = node['refuri'].split('#')[1]
@@ -870,10 +866,18 @@ class ConfluenceWikiTranslator(ConfluenceTranslator):
             label = node.astext()
             for find, encoded in SPECIAL_VALUE_REPLACEMENTS:
                 label = label.replace(find, encoded)
-            if label == doctitle and not anchor:
+            if not doctitle:
+                # A link to an anchor on the current page.
+                assert anchor
+                self.add_text('[%s|%s]' % (label, anchor))
+            elif label == doctitle and not anchor:
+                # A link to a page.
                 self.add_text('[%s]' % label)
             else:
+                # Some text to display and the link is to an anchor in a
+                # different document.
                 self.add_text('[%s|%s%s]' % (label, doctitle, anchor))
+
             raise nodes.SkipNode
 
         # Anchor.
@@ -1007,3 +1011,9 @@ class ConfluenceWikiTranslator(ConfluenceTranslator):
         if 'confluence' in node.get('format', '').split():
             self.add_text(node.astext())
         raise nodes.SkipNode
+
+    def visit_todo_node(self, node):
+        self.visit_admonition(node)
+
+    def depart_todo_node(self, node):
+        self.depart_admonition(node)
--- a/test/unit-tests/common/dataset-common/todo.rst
+++ b/test/unit-tests/common/dataset-common/todo.rst
@@ -1,0 +1,36 @@
+:orphan:
+
+todo test
+---------
+
+This section is going to be
+
+.. todo::
+
+    My first todo entry
+
+losts of ToDo entries so that we can 
+
+.. todo::
+
+    Once I've done that, I need to finish
+    this one as well.
+
+test the Confluence ToDo output and improve the generated Confluence pages.
+
+.. todo::
+
+    And lastly I have this very log todo which should get split
+    nicely across as many lines as is required but will probably be rendered
+    in Confluence WiKi markup as a very long single line.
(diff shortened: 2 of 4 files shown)


sphinxcontrib/confluencebuilder/translator/wiki.py
```diff
@@ -870,10 +866,18 @@ def visit_reference(self, node):
             label = node.astext()
             for find, encoded in SPECIAL_VALUE_REPLACEMENTS:
                 label = label.replace(find, encoded)
-            if label == doctitle and not anchor:
+            if not doctitle:
```
jdknight: While I don't mind mixing related changes in a single PR (in this case, adding more in-depth support for self-page anchors and todo support), I would extremely recommend splitting this up into multiple commits so it's easier to reflect back on changes made when looking at older commit messages.

It appears to me you've removed the top `doctitle` check to help generate anchors on the same document for documents which don't have a title value (side note, title-less documents will never be included in part of the publish event). I believe this introduces two possible issues. First, when `refuri` is set, it should be (assumption) that the reference is to another document (maybe itself) in the doctree. In the event that a document references another document without a title, instead of generating a warning message an invalid anchor reference will be created (e.x. doca to docb#section could generate an anchor link in doca called `[#section]` if docb has no title). Second issue is that in the event a page links to another title-less page without an anchor, `anchor` will be empty and the `assert` will be triggered.

If you really want to permit support anchor references contained within a single document with no title, there should be a check if `docname == self.docname` before making the assumption that an internal anchor will work (maybe only also permitting this case when an anchor value is detected as well).

sphinxcontrib/confluencebuilder/translator/wiki.py
```diff
@@ -1007,3 +1011,9 @@ def visit_raw(self, node):
         if 'confluence' in node.get('format', '').split():
             self.add_text(node.astext())
         raise nodes.SkipNode
+
+    def visit_todo_node(self, node):
```
jdknight: Since both `visit_todo_node` and `depart_todo_node` follow a non-wiki-specific pattern, these could instead be added into `translator/shared.py` instead. It may be nice to also include a comment that this is a `sphinx.ext.todo` hook instead of a standard hook.

test/unit-tests/common/test_common.py
```diff
@@ -13,6 +13,8 @@ class TestConfluenceCommon(unittest.TestCase):
     @classmethod
     def setUpClass(self):
         self.config = _.prepareConfiguration()
+        self.config['extensions'].append('sphinx.ext.todo')
```
jdknight: Since this PR is targeting support for an extension (`sphinx.ext.todo`), I would recommend we create a separate unit testing folder to help separate base features vs. extension features. Instead of placing these tests in `test/unit-tests/common`, it maybe better to place them into `test/unit-tests/ext-todo` instead.

test/unit-tests/common/dataset-common/todo.rst
```diff
@@ -0,0 +1,36 @@
...
+todo test
+---------
+
+This section is going to be
+
+.. todo::
+
+    My first todo entry
+
+losts of ToDo entries so that we can
```
jdknight: Lots?

sphinxcontrib/confluencebuilder/translator/wiki.py
papadeltasierra: I'll see if I can come up with some tests and examples that hit this.  Nothing seemed to hit this code and as per my comment on this PR, I was surprised to find that it was required.  Bit busy with GCSEs and A-levels at the moment (my daughters not mine :-) ) so apologies for not having looked at this sooner.

jdknight:
I'm not certain why this PR was marked closed; possibly when the `next` branch was removed after the release/sync was performed. Free feel to re-submit a new PR with mentioned changes if you have time in the future and I'd be happy to take a look.

jessetan:
@jdknight is the preferred way to add support for extensions to confluencebuilder like in this PR, or to add support for confluencebuilder in the extension itself like in https://github.com/sphinx-contrib/plantuml/pull/23?

jdknight:
@jessetan, I would say it would vary on the specific extension.

This pull request was bringing support for the `todo` extension which is part of the official Sphinx repository (i.e. exists under [`sphinx/ext`][1]). Since it is part of the official repository, I have no objections attempting to maintain support for these extensions from within this extension's implementation (unless there is a compelling reason not to do so).

As for `plantuml`, I would say this is more of an independent extension (even if they are both part of `sphinx-contrib`) and I would not think it would be appropriate to add this information directly into this extension.

There was a pull request in the past (#165) to help expose this extension to other extensions via Sphinx-defined means, which I believe is a correct way to promote extension interaction (if extensions need to find a way to talk). In most cases, I would expect there would be a Sphinx-way to allow various extensions to work with each other. There could be some corner cases where you could want "extensions on an extension" -- while I doubt that is the case, something like that could be achievable, but the use case would be best discussed to see if it actually make sense.

With respect to anything that `plantuml`'s contributors/maintainers have done to interact with this extension, I do not think I would have anything useful to say at this time. I have not investigated on how that extensions works -- I only assume there is a way to convert PlantUML diagrams into images. I would be curious why they would need to have any specific interaction with this extension since one could translate markup into specific images during a transform stage and the Confluence builder would just use the generated image content in the final documentation. Although it is easy just to say that -- the `plantuml` extension could be doing a number of things.

I hope this information helps

[1]: https://github.com/sphinx-doc/sphinx/tree/master/sphinx/ext

jessetan:
Thanks, that clears things up.

Issue #117, opened 2018-07-13
UnknownMacroMigrationException on autodoc and kwargs


MorganGellert:
It seems that the '{' character is not properly being escaped when it is part of the arguments of a function. For instance the code:

```
class Bar(object):
  """ This is a testing class.
  """
  def __init__(self, a, b={'baz': 100}):
    """ Create Bar class object.

    :param a: Slot 1
    :param b: Slot 2 (should fail)
    """
    self._a = a
    self._b = b
```

Fails with the error `com.atlassian.confluence.content.render.xhtml.migration.exceptions.UnknownMacroMigrationException: The macro ''baz'' is unknown.`

This seems to be related to #75. 

Will this be fixed with #51 and be in the 0.1 release?

Thanks!

jdknight:
@MorganGellert, thanks for reporting this. Queued up a pull request which should resolve this issue.

While this should allow you to publish with the example content you've provided, note that autodocs support is still very limited/crude (as you may have already seen in #51).

I'd imagine this fix should be available (after review) in a next possible release; however, if you're keen on having better autodocs supports, it may not be available for a couple of months (see #116).

MorganGellert:
No worries, I just wanted to file the bug. I'm looking forward to autodoc support and plan on using this project in the future. Thanks for all the great work!

jdknight:
Version 1.0 of this extension has been released which includes the storage format issues discussed in this issue; closing.

TxBillbr:
Running 2.0.0 and still seeing this occur.  A specific snippet of code `comment_url = f'{bamboo_url_base}/result/{plan_key}-{build_number}/comment'` causes exception on {build_number}

PR #119, opened 2018-07-15
additional wiki encode corrections

jdknight:
A few locations in the Wiki translator processes user-defined text without encoding in a fashion which isn't save for the conversion-to-storage-format request. One location is within `visit_definition` for definition entries and another exists in `visit_desc_parameter` which is used when processing description lists (e.x. autodoc).

This change corrects #117.


jdknight:
If pull request #121 is approved, this pull request no longer required.

jdknight:
Dropping this change in support for storage-format based pull requests.

PR #120, opened 2018-07-21
publisher: handle various space names in desc-check

jdknight:
When a REST-based publishing event checks for page descendants (when purging legacy pages), a CQL search request can be made when attempting to find descendants for a space. The provided space name is not quoted, and causes the CQL expression to break with some valid space name values (for example, personal spaces which are prefixes with a tilde (`~`) character). Enclosing the space name with quotes prevent an expression error.

### DIFF
--- a/sphinxcontrib/confluencebuilder/publisher.py
+++ b/sphinxcontrib/confluencebuilder/publisher.py
@@ -205,8 +205,8 @@ class ConfluencePublisher():
             if page_id:
                 search_fields = {'cql': 'ancestor=' + str(page_id)}
             else:
-                search_fields = {'cql': 'space=' + self.space_name +
-                    ' and type=page'}
+                search_fields = {'cql': 'space="' + self.space_name +
+                    '" and type=page'}
 
             # Observed issues with "content/{id}/descendant"; using search.
             rsp = self.rest_client.get('content/search', search_fields)


sphinxcontrib/confluencebuilder/publisher.py
```diff
@@ -205,8 +205,8 @@ def getDescendents(self, page_id):
             if page_id:
                 search_fields = {'cql': 'ancestor=' + str(page_id)}
             else:
-                search_fields = {'cql': 'space=' + self.space_name +
-                    ' and type=page'}
+                search_fields = {'cql': 'space="' + self.space_name +
```
tonybaloney: what about if the space name contains a double quote?

jdknight: At any point where `getDescendents` is invoked, the configured space name has already been validated [existence check; and space name character requirement includes `a-Z0-9` or (personal spaces only) `~`].

Since this is the only CQL call in this extension, I do not see a major need to attempt to make a complete CQL escape helper call. That being said, if you are strongly for making it "perfect", I can try to make a proper CQL statement builder utility class.

tonybaloney:
Does this change still apply after #121?

jdknight:
@tonybaloney, yup!

Issue #94, opened 2018-03-07
Embed images on Cofluence page


brxie:
Hi there,

Thanks for making awesome extension!
Could you please let me know, is it in plans to make support for uploading images to Cofluence?
Didn't see it on Supported Markup list.

Regards,
brxie

akall80:
Hello, 
I quote @brxie: wonderful extension, seamlessly working.
Missing images linking and uploading. Do you have any plan?

Many thanks and regards,
A

jdknight:
Unless someone else attempts to provide such a contribution, here's the order of things to which I'll have the opportunity to possibly start image markup support:

1. First, the slew of pending pull requests need to be reviewed and possible accepted (when @tonybaloney has the time; unknown duration).
2. Once the pull requests have been assessed, I have a series of Confluence validation tests changes I'd like to submit and verify existing markup. Ideally from this, generate a new stable version (approximately 1-2 weeks plus release review time).
3. Address the feature request outlined in issue #51 (approximately 1 week to a month).

Once these are addressed, I don't mind attempting to support image markup.

jdknight:
I am going to close this issue in favour of issue #131 (where progress updates on image support will be updated).

PR #121, opened 2018-07-23
storage format conversion (and more)

jdknight:
Here's a big pull request. 😅 

----

Performs a significant change to this extensions lower layer by changing the translation from a Confluence wiki-format document builder to a Confluence storage format builder. The primary reasons for this change are:

- Confluence's primary format is the storage format. This extension, when publishing, will no longer need to perform a two-pass publish event for each document. Originally, documents would need to be converted from a wiki-formatted document to a storage formatted document followed by a second publish event to the Confluence instance.
- Several desired markup changes are limited when using the wiki-format -- tables could not be nested, indentation was limited, and more. With this change, several restructuredText/Sphinx markup have been and can be improved on.

The switch to storage format affects multiple files. The translator has been (mostly) re-written and unit test expected results have all been updated to reflect the new format. During the transition from wiki-format to storage format, some restructuredText/Sphinx markup support has been improved upon (due to the nature of the format changes). A summary of changes can be seen in the `CHANGES.rst` file provided in this pull request. Markup changes have been introduced at a per-commit level, providing more detailed information about the changes when re-introducing a translator hook.

Finally, `README.rst` and documentation has been updated to reflect the changes introduced from this pull request.

You may also reflect on the published validation tests using these changes at the following URL:

> sphinxcontrib-confluencebuilder -- devel
> https://jdknight.atlassian.net/wiki/spaces/confluencebuilder/pages/323463/devel


jdknight:
After completing the last pass of unit tests, validation set and documentation updates, I believe I have completed the series of changes I want to make in part with the storage-format upgrade and dependent stacked changes.

If I have time, I will probably take a gander at #94. Any changes I make will be stacked off these changes, so I will wait for this pull request to be reviewed (and possibly merged) before attempting to submit anything else (so this PR does not get bigger than it already is).

tonybaloney:
@jdknight this PR is really too large to review and provide any meaningful feedback. What I would normally be looking for in something this large is :

* Breaking changes in API
* Changed to behaviour and how that impacts users
* Changes to test exposure/strategy

Great work btw

jdknight:
@tonybaloney, I completely agree that this pull request is extremely large and extremely painful for a review. The switch to storage format just required to changed so much and it was hard to find a way to separate the stack of changes into multiple pull requests. On the plus side, I do not see a feature in the future that would require a massive change such as this again. 🤞

Issue #96, opened 2018-03-19
password in conf.py options


drafter250:
I heard about this project on a Recent Talk Python podcast and was excited to try it out. Thanks for this Project!
In many organisations they use their LDAP passwords to login to confluence wiki. I dont think putting passwords in the conf.py is a good thing to encourage as many people include conf.py in their repo's. I think including a path variable to a password file is a better option no?

jdknight:
@drafter250,

I don't think there's an intent to encourage users to place plain text password (let alone any other encoded, etc. password value) into a committed repository. The configuration and tutorial examples just talk about configuring the value `confluence_server_pass` for a password/API token. Nothing stops an individual from extending their configuration to retrieve a password from either an environment variable or an external file (we actually do [environment variable passing](https://github.com/tonybaloney/sphinxcontrib-confluencebuilder/blob/2d694c5bebab8e53bc50e4d6e6c964d7a5dbf783/test/test_publisher.py#L50) and [external file passing](https://github.com/tonybaloney/sphinxcontrib-confluencebuilder/blob/2d694c5bebab8e53bc50e4d6e6c964d7a5dbf783/test/test_publisher.py#L95-L104) of a password value in our publishing tests).

I do not think the security of the password/API key was attempted to managed by this extension. I don't know if we would want to introduce official environment variables or special files since (I feel) there could be various user scenarios where some authentications work and some do not. If there's a standard/common way to deal with authentication in Python programs that you'd suggest we consider, I'd be willing to take a look (I've only really worked in a limited scope in the Python world).

That being said, I wouldn't be opposed to add additional notes in our documentation to inform the user on not recommending to store plain text/encoded passwords in their own repositories.

nklapste:
I used `getpass.getpass()` in my `conf.py` for creating a simple manual runtime confluence password prompt to fill in `confluence_server_pass` on starting a sphinx build. It may not be the most automated solution, but I found it worked effectively for minimum release projects. This could be a simple (yet limited) recommendation to implement into a `conf.py` for not using plain text passwords in projects.

GhostofGoes:
Doing the same as @nklapste in a project, works fine as a workaround.

digulla:
I think a simple example how to load a user/password combination from a file in the user's home folder would be great.
Using env variables is a security risk; they are visible to everyone on the same computer.

tonybaloney:
I can add an example for a credential store like Hashicorp Vault?

[...]

PR #128, opened 2018-07-29
provide a password prompt option

jdknight:
Introduces a new option `confluence_ask_password` to allow a user to request this extension to prompt for a password to be used for authenticating with a publish event to the configured Confluence instance.

This feature is to assist in user's who wish to invoke a publish request on a documentation set without having to manually adjust the documentation's configuration, setup external environment/files or provide a clear password in an interactive session. With this feature, user's can invoke a build with the following option to a password prompt to appear:

```
sphinx-build [options] -D confluence_ask_password=1 <srcdir> <outdir>
```

This commit explicitly leaves the `getpass` prompt value as empty and provides the prompt information using system write calls. This is to provide more feedback for Cygwin/MinGW sessions (invoked without `winpty`) which cannot handle pulling a password from `getpass` (instead of the session blocking with no input provided).

This change has been introduced based off comments from issue #96.

### DIFF
--- a/sphinxcontrib/confluencebuilder/builder.py
+++ b/sphinxcontrib/confluencebuilder/builder.py
@@ -15,11 +15,13 @@ from .util import ConfluenceUtil
 from .writer import ConfluenceWriter
 from docutils import nodes
 from docutils.io import StringOutput
+from getpass import getpass
 from os import path
 from sphinx import addnodes
 from sphinx.builders import Builder
 from sphinx.util.osutil import ensuredir, SEP
 import io
+import sys
 
 # Clone of relative_uri() sphinx.util.osutil, with bug-fixes
 # since the original code had a few errors.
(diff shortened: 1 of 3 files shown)


tonybaloney:
please resolve conflicts in doc/configuration.rst

sphinxcontrib/confluencebuilder/builder.py
```diff
@@ -70,6 +72,16 @@ def init(self, suppress_conf_check=False):
         if not ConfluenceConfig.validate(self.config, not suppress_conf_check):
             raise ConfluenceConfigurationError('configuration error')
 
+        if self.config.confluence_ask_password:
+            print('(request to accept password from interactive session)')
+            print(' Instance: ' + self.config.confluence_server_url)
+            print('     User: ' + self.config.confluence_server_user)
+            sys.stdout.write(' Password: ')
+            sys.stdout.flush()
+            self.config.confluence_server_pass = getpass('')
```
digulla: Are these three lines working around a bug in getpass()? If so, how about moving this to a helper method with proper documentation before the next guy just "fixes" it to `getpass('Password:')`

PR #135, opened 2018-08-11
asset (images and download) support

jdknight:
Providing the initial implementation to support asset management between a user's Sphinx documentation set and a Confluence instance's attachments. The pull request focuses on supporting Sphinx's `download` directive \[1\] and reStructuredText's images directives (`image` and `figure`) \[2\].

## asset management
A simple overview of asset management is as follows:

- A document is scanned for assets that will be used in documents.
- Found assets are tracked (file path, hashes, etc.) inside an asset manager.
- When a translator parses a document with assets, it will refer to the asset manager on information on how to form a reference to the future Confluence attachment.
- When publishing:
  - A capture of all known assets in a Confluence instance is tracked (i.e. possible legacy assets).
  - Assets are published to specific documents as attachments.
  - Once all assets are published, legacy assets (if any) are removed.

The processing and tracking of assets is managed inside the `ConfluenceAssetManager` implementation.

## hashing for duplication and re-publish support
Detected assets will have a calculated hash value. This hash value serves two purposes:

 1. A duplicate asset check. A document can define the path to an asset in various ways (relative, absolute) and may even (oddly) have multiple files of the asset. The manager has the ability to check if two (or more) files have the same hash. If so, only a publish event is needed (and each document will reference the same attachment).
 2. Limit re-publishing. When an asset is stored as an attachment in Confluence, the publish event will store the hash information inside the attachment's comment. If a user performs a publish event again with the same documentation set, an attachment's comment will be used to check if the hash value is changed. If the hash value has not changed, the attachment will not be re-published.

## download directive
Downloads can be visualized in two ways. First, if the document does not provide an explicit label for the directive, Confluence's "view-file" macro will be used. Second, if an explicit label is provided, the download reference will use a simple hyperlink (with the label value as the link text).

## image directives

Support is added for image URIs include external references, internal references and embedded images. External images will build an image markup pointing to the provided URI. Internal images will refer to the asset manager on where to find a future attachment resource. Embedded images need to be supported via a quirk. The Sphinx engine provides the ability to convert an embedded image (i.e. an image defined by a data URI schema) into a generated image resource. While nice, this only happens are a post-transformation stage per document (which our asset implementation desires to know all asset content before attempting any translations). To workaround this, asset information is re-checked for new entries after a document's doctree is resolved. Any externally generated images will be new/independent images -- even if a generated image contains a hash matching an existing asset, it will be published as its own asset (to prevent complications with sharing assets between multiple documents).

## other notes
Some additional notables:

- Typically a page with uses an asset will be the page which will hold the attachment; however, if multiple pages use the same asset, the master document \[3\] will be used to host the attachment.
- The configuration option `confluence_asset_override` will allow a user for force certain asset management. By default, with a value of `None`, automatic asset management is in play. If the option is configured to `True`, this extension will force publish assets each time. If the option is configured to `False`, this extension will not attempt to publish any assets.
- This pull request includes respective unit test and validation test additions.

## testing

These new changes have been tested using both Confluence Cloud and Confluence Server (v6.0.6) -- both variants appear to be working well.

The newly supported directives can be observed on the following pages (sphinxcontrib-confluencebuilder -- devel):

- [devel-download](https://jdknight.atlassian.net/wiki/x/bYDiK)
- [devel-images](https://jdknight.atlassian.net/wiki/x/zYDfK)

\[1\]: http://www.sphinx-doc.org/en/stable/markup/inline.html#role-download
\[2\]: http://docutils.sourceforge.net/docs/ref/rst/directives.html#images
\[3\]: http://www.sphinx-doc.org/en/master/glossary.html#term-master-document


meltingturret:
nice job working great on confluence but when running HTML build I get:

`$ sphinx-build -b html . _build/html -E -a`

`Exception occurred:
  File "c:\python27\lib\site-packages\sphinxcontrib\confluencebuilder\__init__.py", line 34, in assetsDocTreeResolvedHook
    app.builder.assets.processDocument(doctree, docname, True)
`

apenney:
Just wanted to say thank you for this!  I just realized today, while evaluating confluencebuilder that images don't just magically work, and then stumbled over this in the PR section.  I definitely appreciate all the work you put into this and hope it gets merged. :)

jdknight:
@meltingturret, thanks for giving it a spin and finding this issue. Made a correction to prevent the custom `doctree-resolved` hook to only occur for this extensions builder. Hopefully there should be no conflict with other builders.

PR #139, opened 2018-08-14
translator: add interim math handling

jdknight:
Add a series of node translation hooks for Sphinx's `math` capability, This is to prevent a hard failure (`unknown_visit`) if a user's document contains math-related directives.


tonybaloney:
Should also add a note in docs that this is unsupported

PR #142, opened 2018-08-15
manpage support

jdknight:
Adds support for Sphinx's manpage role \[1\]. When a manpage URL is defined in the documentation set's configuration, an external reference will be built for the manual page's path; otherwise a simple emphasis of the manpage type will be presented (in the same fashion as Sphinx's HTML builder).

\[1\]: http://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html?highlight=manpage#role-manpage


tonybaloney:
Please rebase

Issue #144, opened 2018-09-06
autodoc: don't work links to modules in table of contents and links to custom types


skudin:
Hello.

I tried generate documentation for my package by autodoc. HTML version generated successfully without warnings, but in confluence version don't works links to modules in table of contents on starting page and links to custom types, e.g. custom exceptions in desribe of method.

While execution of make.bat confluence, I got warnings like those: 
WARNING: unable to build link to document due to missing title (in index): genindex
WARNING: unable to build link to document due to missing title (in index): py-modindex
WARNING: unable to build link to document due to missing title (in index): search
WARNING: unable to build link to document due to missing title (in mypackage.mymodule): mypackage.mymodule.conf#mypackage.mymodule.exceptions

My enviroment is Windows 10 Pro, Python 3.6.4, Sphinx 1.7.8 and sphinxcontrib-confluencebuilder (main branch, current version, last commit is 330e31af).

verdesmarald:
I have the same issue, not sure about genindex/search but the problem with the other links seems to be that `translator._visit_reference_intern_uri` not strip the anchor portion before calling `ConfluenceState.title`.

I'll try and fix this and put in a PR.

jdknight:
I suspect `genindex`/`py-modindex`/`search` are the default generated pages from `sphinx-quickstart` -- which these documents are not generated with a title, thus will never be published. I assume though that this is not the issue you are concerned about.

It is most likely the issue that @verdesmarald has mentioned (and submitted a pull request for). Hopefully this change will resolve your issue.

While this extension's `master` branch now lists autodoc as supported, clearly there is still limitations in the initial support. I will hopefully be able to add more additional examples to test against in the future. Also, if anyone knows of some open source projects with a great sample set of documentation, I would not mind doing some tests against these source sets to help correct other issues that may exist.

jdknight:
Version 1.0 of this extension has been released which addresses the anchor issues observed with autodoc content. While other autodoc changes can be made, they are currently being tracked in other active issues. Closing this issue.

PR #147, opened 2018-09-20
Strip anchor before trying to determine doctitle

verdesmarald:
I had the same issue as #144, this change seems to fix the issue for internal links with anchors.

### DIFF
--- a/sphinxcontrib/confluencebuilder/translator.py
+++ b/sphinxcontrib/confluencebuilder/translator.py
@@ -821,7 +821,7 @@ class ConfluenceTranslator(BaseTranslator):
 
     def _visit_reference_intern_uri(self, node):
         docname = posixpath.normpath(
-            self.docparent + path.splitext(node['refuri'])[0])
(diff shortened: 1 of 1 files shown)


jdknight:
@verdesmarald, thanks for the PR. The fix looks good; however, you mind re-writing your commit message with proper sign-off message (see [CONTRIBUTING.rst](https://github.com/tonybaloney/sphinxcontrib-confluencebuilder/blob/master/CONTRIBUTING.rst))?

As for why the build breaks, I can look into that. Something's up with Sphinx 1.8 and the highlight directive.

verdesmarald:
> @verdesmarald, thanks for the PR. The fix looks good; however, you mind re-writing your commit message with proper sign-off message (see [CONTRIBUTING.rst](https://github.com/tonybaloney/sphinxcontrib-confluencebuilder/blob/master/CONTRIBUTING.rst))?
> 
> As for why the build breaks, I can look into that. Something's up with Sphinx 1.8 and the highlight directive.

@jdknight Updated the commit message now, sorry I was away overseas for a fortnight! 😄

PR #154, opened 2018-11-09
builder: handle ref with no id in remove-title-check

jdknight:
When a configuration flags to remove the initial title from Confluence pages, additional work is performed to remove local references from those titles so that a local table of contents does not generate a bad link to itself. This check compares the references identifiers to the matching title's reference identifier before removing; however, not all references may have identifiers (not a requirement for docutils). Adjusting to ensure we check if any identifiers exist before pulling an identifier to check against.


tonybaloney:
Can you refactor that line please. Having to put parenthesis around if-evaluation statements is normally a sign for refactor

jdknight:
Can do; I’m unavailable until December so I can get to it then.

PR #163, opened 2018-12-20
Make '-Dconfluence_publish=...' work

jhermann:
When passed on the command line, options are strings. Make a few
common negative strings known to 'confluence_publish'.

### DIFF
--- a/sphinxcontrib/confluencebuilder/builder.py
+++ b/sphinxcontrib/confluencebuilder/builder.py
@@ -124,7 +124,8 @@ class ConfluenceBuilder(Builder):
         else:
             self.lang_transform = None
 
-        if self.config.confluence_publish:
+        if self.config.confluence_publish and (self.config.confluence_publish.lower()
+                                               not in ('0', 'false', 'n', 'no', 'off')):
             self.publish = True
             self.publisher.connect()
         else:
--- a/sphinxcontrib/confluencebuilder/builder.py
+++ b/sphinxcontrib/confluencebuilder/builder.py
@@ -124,7 +124,7 @@ class ConfluenceBuilder(Builder):
         else:
             self.lang_transform = None
 
-        if self.config.confluence_publish and (self.config.confluence_publish.lower()
+        if self.config.confluence_publish and (str(self.config.confluence_publish).lower()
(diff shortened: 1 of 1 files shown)


jhermann:
BTW, this logic could be applied to every boolean config switch.

jdknight:
@jhermann, I will agree that there is some limitations when interpreting non-string configuration entries provided via the command line. For example, I would expect that if a project's configuration registers `confluence_publish=True` that the following invoke will prevent publishing:

```
sphinx-build -b confluence . _build/confluence -Dconfluence_publish=False
```

However, as mentioned in this pull request, the string `False` will evaluate to `True` for the publish condition check -- which I believe (as well as most people may believe) is counter-intuitive. While not trying to push off a change, this issue also exists with several boolean-based configuration entries in Sphinx as well. It will be nice to cleanup the configuration parsing, but I first want to mention that a user can workaround overriding boolean configuration options by not specifying any value:

```
sphinx-build -b confluence . _build/confluence -Dconfluence_publish=
```

Back to the change request, I personally not a fan of trying to interpret all the proposed string types to check for the boolean's desired state. I am for checking against `false` since it is consistent with Python's boolean pattern of "True/False", and I lean towards liking "0" as it is a much easier/lazier option to apply when using the command line. I am less fond of trying to interpret "n[o]/off" since I believe it would be better to only promote the Python's boolean value types ("True/False"). That all being said, maybe I am being to strict about this. I do not plan to deny a change based on this since it makes configuration parsing easier for users; however I would rather make sure we do not attempt to promote the use of setting boolean values to "Yes", "On", etc. is any documentation.

As mentioned in the pull request, this change only addresses `confluence_publish` and not other boolean-based configuration entries. I wonder if there is a better way to address the more-protective configuration interpretation. I do not know off hand, but if configurations are interpreted as soon as `add_config_value` is invoked (in `setup`), maybe values can be checked and overridden so elaborate boolean interpretations do not need to crawl into other portions of the implementation. If this cannot be done inside `setup`, maybe someone in the initial phase of the builder's initialization (before `ConfluenceConfig.validate` is invoked – i.e. introduce a `ConfluenceConfig.finalizer`).

PR #164, opened 2018-12-21
add confluence_publish_postfix / confluence_publish_numbered settings

jhermann:
confluence_publish_postfix is just complementary to
the existing confluence_publish_prefix.

confluence_publish_numbered adds a running counter to each page
title, so they are ordered as they appear in the Sphinx document.
The REST API unfortunately does not offer ordering (tickets to
fix that are open for years now, as usual), so this is the next
best thing.


jhermann:
Example of resulting sidebar hierarchy…
sidebar

jjcaballero:
This pull request seems blocked due to unattended requested changes in the review since 5 months ago. Is there any plan to take over this to unblock this situation? The possible features included here are really interesting and valuable, it's a pity not being able to use them. They can solve or workaround very common problems related to Confluence (page order and unique page names). Thanks!

PR #167, opened 2019-01-13
translator: ignore enumerated list style if not set

jdknight:
List styling support for enumerated lists checks for a populated `enumtype` value; however, there is no guarantee that docutils will populate the node with a specific enumeration type. If a type is not provided, ignore attempting to apply a style (i.e. use Confluence's default type for ordered lists).

### DIFF
--- a/sphinxcontrib/confluencebuilder/translator.py
+++ b/sphinxcontrib/confluencebuilder/translator.py
@@ -266,24 +266,26 @@ class ConfluenceTranslator(BaseTranslator):
         #         'start' attribute of an ordered list tag; limiting to
         #         auto-enumeration items only.
         list_style_type = None
-        if node['enumtype'] == 'upperalpha':
-            list_style_type = 'upper-alpha'
-        elif node['enumtype'] == 'loweralpha':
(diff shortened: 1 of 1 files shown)


sphinxcontrib/confluencebuilder/translator.py
```diff
@@ -266,7 +266,9 @@ def visit_enumerated_list(self, node):
         #         'start' attribute of an ordered list tag; limiting to
         #         auto-enumeration items only.
         list_style_type = None
-        if node['enumtype'] == 'upperalpha':
+        if 'enumtype' not in node:
```
tonybaloney: Can't you use `if 'enumtype' in node and node['enumtype'] == 'upperalpha'` instead? It's more explicit why the first clause is there. I'd find the `pass` confusing without knowing about this issue

jdknight: I have tweaked the change to remove the "is None" check, but opted for saving the condition where if no `enumtype` is provided, the `list_style_type` will not be set to avoid any unnecessary styling modifications. I hope that works for you.

PR #169, opened 2019-01-25
Explicitly define a subset of document names to publish

bosonogi:
If subset of document names is defined using the `confluence_doc_subset` option, only those documents are processed and uploaded.

This is especially useful for large projects with demanding plugins for building pages (like `sphinxcontrib-plantuml` for example). When starting a build from scratch or when publishing every page is not desired, it is good to be able to explicitly control what will be done.

Using this options implicitly disables `confluence_purge` as keeping it enabled would delete every document that is not in the subset.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -567,24 +567,6 @@ configuration).
 
    confluence_proxy = 'myawesomeproxy:8080'
 
-.. _confluence_timeout:
-
-confluence_timeout
-~~~~~~~~~~~~~~~~~~
-
-Force a timeout (in seconds) for network interaction. The timeout used by this
-extension is not explicitly configured (i.e. managed by Requests_ and other
-implementations). By default, assume that any network interaction will not
-timeout. Since the target Confluence instance is most likely to be found on an
-external server, is it recommended to explicitly configure a timeout value based
-on the environment being used. For example, to configure a timeout of ten
-seconds, the following can be used:
-
-.. code-block:: python
-
-   confluence_timeout = 10
-
-
 .. |confluence_publish_subset| replace:: ``confluence_publish_subset``
 .. _confluence_publish_subset:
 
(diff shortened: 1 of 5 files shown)


bosonogi:
@jdknight I'm ashamed to admit that I did not know about that sphinx-build option for processing individual files.

This is what I found after a quick trial of using this option: building worked as expected, but publishing gave warnings about missing files.

I'll look into it more to see if there is any benefit to what I did here.

bosonogi:
@jdknight I've opened a separate PR to deal with the publishing issue: #173

One difference between Spihinx's subset feature and the one implemented here is that Sphinx will additionally process document dependencies. This is the correct thing to do in general obviously, but if you have say a document with many dependencies deep in your document hierarchy; fixing a typo in a paragraph for example would re-build and re-publish a bunch of pages unnecessarily.

In these cases the `confluence_doc_subset` option could prove useful as a tool to more finely control what is being done. It gives you explicit control for when you know what you are doing, ie. when you know processing certain document dependencies can be omitted.

bosonogi:
Thank you for the detailed feedback @jdknight 
I tweaked this pull request to address your notes.

doc/configuration.rst
```diff
@@ -584,6 +584,52 @@ seconds, the following can be used:
 
    confluence_timeout = 10
 
+
+.. |confluence_publish_subset| replace:: ``confluence_publish_subset``
+.. _confluence_publish_subset:
+
+confluence_publish_subset
```
jdknight: Mind if I do some of my own tweaks for the proposed documentation changes (and stack them in this pull request)?

bosonogi: I don't mind @jdknight 
tweak away :)

jdknight:
@bosonogi, have any comments about the documentation changes I have submitted?

doc/tips.rst
```diff
@@ -86,6 +86,35 @@ may not desire two failed attempts when publishing a documentation set. To
...
+
+Users have the ability to publish a subset of processed documents by using the
+``confluence_publish_subset`` (:ref:`jump<confluence_publish_subset>`) option.
+This can be useful for large documentation sets where a user may wish to only
+publish an update for one or more documents (instead of the entire set). While
+subset publishing can be commonly used by setting the
+``confluence_publish_subset`` option in a command line build, this may not be
+ideal for some environments. The following is a code snippet, which when
+included in a project's ``conf.py`` file, will provide a means for a user to
+specify a file outlining which documents are desire:
```
bosonogi: @jdknight typo: "desire"

bosonogi:
@jdknight Good job with the doc changes, thank you.

jdknight:
@bosonogi, thanks for your contribution; much appreciated.

PR #172, opened 2019-02-04
Support Sphinx's parallel mode (the `-j` option to `sphinx-build`)

bosonogi:
Without these changes `sphinx-build` refuses to work in parallel.

In my projects, using the `-j` option with these changes did not produce any ill effects. Is there a reason it shouldn't be allowed?

### DIFF
--- a/sphinxcontrib/confluencebuilder/builder.py
+++ b/sphinxcontrib/confluencebuilder/builder.py
@@ -52,9 +52,9 @@ def relative_uri(base, to):
 
 
 class ConfluenceBuilder(Builder):
+    allow_parallel = True
     name = 'confluence'
     format = 'confluence'
-    allow_parallel = True
 
     def __init__(self, app):
         super(ConfluenceBuilder, self).__init__(app)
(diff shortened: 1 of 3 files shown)


sphinxcontrib/confluencebuilder/builder.py
```diff
@@ -24,6 +24,7 @@
 import io
 import sys
 
+
```
jdknight: (optional) While I believe you are updating the code to be more PEP 8 compliant, this change is not specifically related to this commit. I would normally suggest if you want to make changes like this is to put them in a separate commit; however, the change is trivial enough that it should not prevent consideration for this pull request.

sphinxcontrib/confluencebuilder/builder.py
```diff
@@ -49,9 +50,11 @@ def relative_uri(base, to):
         return '.' + SEP
     return ('..' + SEP) * (len(b2)-1) + SEP.join(t2)
 
+
 class ConfluenceBuilder(Builder):
     name = 'confluence'
     format = 'confluence'
+    allow_parallel = True
```
jdknight: (optional) I would recommend to sort this attributes lexicographically.

Issue #176, opened 2019-02-25
sphinxcontrib-confluencebuilder request error


keesj:
Hi,

I am trying upload a page with an image in it. Somehow this normally works but somehow it currently fails. The image in question is not special and I tried renaming the image or replacing the contents of the image but the problem remains.

After the upload the page contains an image that looks like this
![selection_230](https://user-images.githubusercontent.com/57112/53331874-42795600-38f2-11e9-8b31-2060e6fc2be7.png)

The input rst contains something like
``
.. image:: img/ts3.jpg
``

I am using the git version of the software and confluence Confluence 6.12.2

[log.txt](https://github.com/tonybaloney/sphinxcontrib-confluencebuilder/files/2900018/log.txt)

```
writing output... [100%] index                                                                                                                             
publishing documents... [100%] electrical                                                                                                                  
publishing assets... [ 40%] img_ts3.png                                                                                                                    
Exception occurred:
  File "/home/why/.local/lib/python3.6/site-packages/requests/models.py", line 159, in _encode_files
    fdata = fp.read()
AttributeError: 'bool' object has no attribute 'read'
The full traceback has been saved in /tmp/sphinx-err-xru26n0a.log, if you want to report the issue to the developers.
Please also report this if it was a user error, so that a better error message can be provided next time.
A bug report can be filed in the tracker at <https://github.com/sphinx-doc/sphinx/issues>. Thanks!
Makefile:20: recipe for target 'confluence' failed
make: *** [confluence] Error 2
```

keesj:
More debugging revealed something new. the error occurs when confluence_disable_notifications = True. In such situations the key value pair given to requests/models contains "minorEdit" "True" and this can not be handled with the request module. e.g. it looks like when uploading a file the minor edit is causing problems

```
    fdata = fp.read()
AttributeError: 'bool' object has no attribute 'read'
> /home/why/.local/lib/python3.6/site-packages/requests/models.py(159)_encode_files()
-> fdata = fp.read()
(Pdb) list
154  	                fp = v
155  	
156  	            if isinstance(fp, (str, bytes, bytearray)):
157  	                fdata = fp
158  	            else:
159  ->	                fdata = fp.read()
160  	
161  	            rf = RequestField(name=k, data=fdata, filename=fn, headers=fh)
162  	            rf.make_multipart(content_type=ft)
163  	            new_fields.append(rf)
164  	
(Pdb) fp
True
(Pdb) p k
[...]

PR #186, opened 2019-03-07
publisher: handle asset publishing with older requests

jdknight:
Confluence's API can accept attachment comments and the `minorEdit` argument through a set of uploaded files [1]. While in the most recent version of requests this is not an issue, requests pre-v2.19 has issues. The `minorEdit` value is provided in boolean form, which the older versions of requests does not handle well. Since this extension still attempts to support at least requests v2.14.0, a workaround is being applied. Instead of configuring the flag with a boolean value, set the value in a compatible string form.

\[1\]: https://docs.atlassian.com/atlassian-confluence/REST/6.6.0/#content/{id}/child/attachment-createAttachments
\[2\]: https://github.com/kennethreitz/requests/commit/8546a15587b2a240c1d1404d3ef99365b911bf13

### DIFF
--- a/sphinxcontrib/confluencebuilder/publisher.py
+++ b/sphinxcontrib/confluencebuilder/publisher.py
@@ -461,7 +461,8 @@ class ConfluencePublisher():
                 }
 
                 if not self.notify:
-                    data['minorEdit'] = True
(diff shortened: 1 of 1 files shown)


jdknight:
Corrects issue #176.

PR #199, opened 2019-05-07
Allow cookie for authentications

mkunzmann:
Allow usage of a cookie for authentication to the confluence API. This
might be useful if the confluence instance is hidden behind some single-
sign-on services.

Signed-off-by: Michael Kunzmann <michael.kunzmann@gmx.net>

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -51,6 +51,8 @@ username value should be used:
 
    confluence_server_pass = 'myawesomepassword'
 
+See also :ref:`advanced authentication options<confluence_advanced_conf_auth>`.
+
 .. caution::
 
    It is never recommended to store an API token or raw password into a
@@ -94,16 +96,6 @@ If using Confluence instance, this value will most likely be the username value.
    confluence_server_user = 'myawesomeuser@example.com'
        (or)
    confluence_server_user = 'myawesomeuser'
-
-confluence_server_cookies
-~~~~~~~~~~~~~~~~~~~~~~~~~
-
-Cookie(s) that can used to authenticate with the Confluence instance. This needs to
-be a dictionary.
-
-.. code-block:: python
-
-   confluence_server_cookies = {'SESSION_ID': 'random session id string', 'U_ID': 'some_user'}
 
 .. |confluence_space_name| replace:: ``confluence_space_name``
 .. _confluence_space_name:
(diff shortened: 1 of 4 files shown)


jdknight:
@mkunzmann, thanks for the enhancement. I am hoping to try this out before merging in ~ but wanted to mention that the changes you have provided look good.

jdknight:
@mkunzmann, out of curiosity, do you use cookie authentication through a proxy/third-party configuration? From what I can tell, Atlassian has been removing their support for cookie-based authentication (promoting API keys instead). This makes me hesitate to add this feature into this extension for a short-lived authentication manner. However, if you have a use case where you are authenticating using other means, I would not mind going ahead and supporting passing session values in REST requests.

On another note, I am wondering if it would be better to tweak all this to support configuring session cookies outside the context of just password management (i.e. remove the implication that `confluence_server_cookies` must be login-related, remove the restrictions of setting a password and cookie information at the same time and always apply cookie information when configured. That way, if a user is interacting with any Confluence instance which has session requirements, a user has freedom to configure respective values they may need. Thoughts?

mkunzmann:
@jdknight Thanks for your response and sorry for my late reply.

I currently use this with cookies for a SiteMinder SSO solution that replaces the JIRA authentication for us. So setting the cookies allows a login, password and username are not required in this case.

If there are use cases where setting both user/password and cookies the restriction can be removed.

jdknight:
_(template-like response)_ Changes in this pull request look to be in good state. Delays in other features in this project will result in this feature being delayed (see [this comment](https://github.com/sphinx-contrib/confluencebuilder/pull/215#issuecomment-541268554) for more details), so the expected window for this feature is planned around end-of-year (2019). Feel free to make additional changes/etc. to this pull request before the planned window. Critical changes may occur on master before this pull request is merged in -- while you are free to help synchronize features as conflicts are detected, I do mind help correcting conflicts myself when I have the opportunity to merge this change in. If you wish to review changes I may stack on this pull request set, feel free to indicate so. When the window to merge this in this pull requests occurs and changes are needed, I can stack on change and wait for your additional approval. Otherwise, I will assume I may stack changes (if needed) which is deemed best for this extension for a quicker integration into master. Labeling this pull request as "planned-end-of-year".

A quick review of this pull request include the following:
- I am hoping to setup a SSO solution to help provide an independent test of this feature (although still expecting the pull implementation will just work).
- Will most likely not tweak to allow both `confluence_server_pass` and `confluence_server_cookies` at the same time.

jdknight:
Finally merged in -- sorry for the delay. Thanks for change.

Issue #202, opened 2019-05-10
Unit tests fail due to changes in Sphinx 2


verdesmarald:
There are at least 2 breaking changes in Sphinx 2 that cause unit test failures:

1. The default value for `master_doc` is now `index` rather than `contents`. This causes one test to fail and several other tests to emit a warning (`WARNING: Since v2.0, Sphinx uses "index" as master_doc by default. Please add "master_doc = 'contents'" to your conf.py.`). This can be easily fixed by adding `config['master_doc'] = 'contents'` to the `prepareConfiguration` method.
1. The default alignment for tables and figures is now centered instead of left. This causes one of the tests to fail because it contains a figure with no `:align:` property. This is more difficult to fix since the tests currently don't cater for differing output depending on the Sphinx version being used.

jdknight:
@verdesmarald, thanks for the notice. With #204 merged in, hopefully this puts this extension in a better state with the most recent (stable) Sphinx release. Will mark as closed.

PR #204, opened 2019-06-19
maintenance for sphinx 2.x support

jdknight:
Provides a series of corrections to the builder, translator and test implementations to support Sphinx variants from v1.6.3 to the most recent v2.1.x series. This includes:

- Handling Sphinx v2+ which defaults tables/figures to be center aligned.
- Handling Sphinx v2+ default `master_doc` change from `contents` to `index`.
- Handling various domain/label assignments adjusted in Sphinx v2.1+.
- Adjusting test logic to explicitly test Sphinx v1.8, v2.0 and v2.1 series.


jdknight:
See also #202.

PR #214, opened 2019-06-26
support previous/next navigation

jdknight:
Provides initial support to have previous/next navigational links at various locations of a document based on a user's configuration. The option `confluence_prev_next_buttons_location` (default disabled with `None`) can be configured with the following to respectively render previous/next buttons for a detected ordered list of documents: bottom, both, top, None.


jacobdr:
This is a useful feature -- would be great to see it merged in

PR #215, opened 2019-07-07
Addresses #143 (Custom directives to support the JIRA structure macro)

cstarner:
Adds a jira and jira_issue directive to act as a wrapper to the JIRA
structured macro available in Confluence.

Allows for configuring sets of named JIRA servers, not providing servers
at all, and a combination of named servers and overriding their
respective values on the directives.

Provides support for both single issue links (via jira_issue) and JQL
links (via jira). Allows for specifying the parameters available to
the macro.

Allows the directives to be disabled via the 'confluence_adv_restricted_macros'
configuration value

Signed-off-by: Colin Starner <colin.starner@gmail.com>

PR #217, opened 2019-07-09
Add image files ending with star (*) instead of specifying the format

jjcaballero:
**Some context:**
I was trying to include plots generated by matplotlib.sphinext when using the confluence builder. Then I learned that plots were properly generated and the only problem was the matplotlib.sphinext extension was not considering confluence as a possible builder.
Therefore, I created this PR (https://github.com/matplotlib/matplotlib/pull/14681) there, to ask them to include it. Eventually, this was addressed in other PR (https://github.com/matplotlib/matplotlib/pull/14683) which enhance the same but in a more general way. Now, builders which want to include plots from matplotlibs should handle images ending with star. 

And handling image with star is the goal of this PR. I have added the parameter **supported_image_types** to the builder like here:  [sphinx builders doc](https://www.sphinx-doc.org/en/master/usage/builders/index.html#sphinx.builders.html.StandaloneHTMLBuilder.supported_image_types) 
We might want to include more image formats, I have just considered png, jpeg and pdf.

**In a nutshell:** 
when the image ends with star and does not exists, then the ConfluenceAssetManager will use the first found image with the same name but ending with a supported image type.

**Side note:** maybe I am wrong, but I think there are no tests to verify the publishing itself.  Is there any recommendation to do such a thing? It would be very convenient to be able to run tests choosing confluence server and parent page. I mean using the same infrastructure used to run tests but publishing.

### DIFF
--- a/sphinxcontrib/confluencebuilder/builder.py
+++ b/sphinxcontrib/confluencebuilder/builder.py
@@ -307,7 +307,9 @@ class ConfluenceBuilder(Builder):
                         if not mf:
                             continue
 
-                        new_node = nodes.image(uri=path.join(self.outdir, mf))
+                        new_node = nodes.image(
+                            candidates={'?'},
+                            uri=path.join(self.outdir, mf))
                         if not isinstance(node, nodes.math):
                             new_node['align'] = 'center'
                         node.replace_self(new_node)
--- a/sphinxcontrib/confluencebuilder/builder.py
+++ b/sphinxcontrib/confluencebuilder/builder.py
@@ -66,6 +66,7 @@ class ConfluenceBuilder(Builder):
     name = 'confluence'
     format = 'confluence'
     supported_image_types = StandaloneHTMLBuilder.supported_image_types
+    supported_remote_images = True
 
     def __init__(self, app):
         super(ConfluenceBuilder, self).__init__(app)
--- a/sphinxcontrib/confluencebuilder/builder.py
+++ b/sphinxcontrib/confluencebuilder/builder.py
@@ -20,6 +20,7 @@ from getpass import getpass
 from os import path
 from sphinx import addnodes
 from sphinx.builders import Builder
+from sphinx.builders.html import StandaloneHTMLBuilder
 from sphinx.errors import ExtensionError
 from sphinx.locale import _
 from sphinx.util import status_iterator
@@ -64,6 +65,7 @@ class ConfluenceBuilder(Builder):
     allow_parallel = True
     name = 'confluence'
     format = 'confluence'
+    supported_image_types = StandaloneHTMLBuilder.supported_image_types
 
     def __init__(self, app):
         super(ConfluenceBuilder, self).__init__(app)
@@ -255,6 +257,7 @@ class ConfluenceBuilder(Builder):
             doctitle = ConfluenceState.registerTitle(docname, doctitle,
                 self.config.confluence_publish_prefix,
                 self.config.confluence_publish_postfix)
+
             if docname in docnames:
                 # Only publish documents that Sphinx asked to prepare
                 self.publish_docnames.append(docname)
@@ -310,6 +313,9 @@ class ConfluenceBuilder(Builder):
                     except imgmath.MathExtError as exc:
                         ConfluenceLogger.warn('inline latex {}: {}'.format(
                             node.astext(), exc))
+
+            # for every doctree, pick the best image candidate
+            self.post_process_images(doctree)
 
         # Scan for assets that may exist in the documents to be published. This
         # will find most if not all assets in the documentation set. The
--- a/test/unit-tests/common/test_image_star.py
+++ b/test/unit-tests/common/test_image_star.py
@@ -1,0 +1,92 @@
+# -*- coding: utf-8 -*-
+"""
+    :copyright: Copyright 2017-2019 by the contributors (see AUTHORS file).
+    :license: BSD-2-Clause, see LICENSE for details.
+"""
+
+import os
+import re
+import shutil
+import tempfile
+import unittest
+
+from sphinxcontrib.confluencebuilder.state import ConfluenceState
+from sphinxcontrib_confluencebuilder_util import ConfluenceTestUtil as _
+from sphinxcontrib.confluencebuilder.builder import ConfluenceBuilder
+
+
+class TestImageStar(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        test_dir = os.path.dirname(os.path.realpath(__file__))
+        cls.image_file = os.path.join(test_dir, 'assets', 'image01.png')
+        cls.image_basename = cls.file_basename(cls.image_file)
+        cls.datasetdir = tempfile.mkdtemp()
+
+    @classmethod
+    def tearDownClass(self):
+        shutil.rmtree(self.datasetdir)
+
+    @staticmethod
+    def file_basename(filepath):
+        return os.path.basename(filepath).rsplit('.')[0]
+
+    @staticmethod
+    def file_extension(filepath):
+        return os.path.basename(filepath).rsplit('.')[1]
+
+    def build_document_given_dataset(self, dataset):
+        config = _.prepareConfiguration()
+        doc_dir, doctree_dir = _.prepareDirectories('image-star')
+        _.buildSphinx(dataset, doc_dir, doctree_dir, config)
+        return doc_dir
+
+    def write_index_file(self, indexfile):
+        with open(indexfile, 'w') as out:
+            out.write('.. image:: {}.*'.format(self.image_basename))
+
+    def write_expected_index_file(self, outfile, actual_extension):
+        ''' writes the expected content into outfile using
+        the actual extension
+        '''
+        image_filename = self.image_basename + '.' + actual_extension
+        content = '<ri:attachment ri:filename="{}">'.format(image_filename)
+        content += '</ri:attachment>\n'
+        content = '<ac:image>\n' + content + '</ac:image>'
+        with open(outfile, 'w') as out:
+            out.write(content)
+
+    def test_image_ending_with_star(self):
+        ''' Test how Confluence Builder resolves the image path when
+        it ends with star (*).
+        A temporary datasetdir is created to store::
+        - image with the extension as the image type to be tested (png): copy
+          of the image01.png from assets folder
+        - index.rst: created on the fly using star notation
+        - expected output file: expected_index.conf (created on the fly too)
+
+        Sphinx builders finds the png as the candidate. Similar tests
+        can be created for all ConfluenceBuilder.supported_image_types
+        but it requires an actual image of each image type.
+        '''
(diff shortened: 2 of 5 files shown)


jdknight:
To start off, it is most likely that #228 will be merged in before this change (which will cause conflicts; but I do not mind correcting these conflicts if needed).

I have no problem with the idea of supporting a wildcard to map to detected assets which may be generated by other extension. I would like to also try this out to actually see it being performed (even though the idea makes sense after reading the referenced PRs). I am a bit hesistent in having this `supported_image_types` list though. Sphinx does have some mime definitions inside their `images.py` implementation -- I wonder if something would be doing with this to prevent this extension to trying to manage supported image types (although, I have no problem definition a separate independent list to whitelist or blacklist image types based off official documented Confluence supported image types.

Note that due to other issues being addressed in this extension,  the expected window for this feature is planned around end-of-year (2019; see [this comment](https://github.com/sphinx-contrib/confluencebuilder/pull/215#issuecomment-541268554) for more details).

jessetan:
I have a similar usecase. I build both HTML and PDF output from a single source. To make images (especially diagrams) display as crisp as possible, I use SVG diagrams for HTML output and PDF diagrams for inclusion into PDF output.
The image is included using `.. figure:: images/diagram.*`. The actual file (`diagram.svg` or `diagram.pdf`) that is included depends on the builder that is used.

Sphinx already has code that handles an asterisk in the file extension: https://github.com/sphinx-doc/sphinx/blob/01f8dad160d206a48302b1bcd91b4e0bc36e8a10/sphinx/builders/__init__.py#L186. This replaces the `*` in the filename based on the `supported_image_types` of a writer.
This is called by the default writers in Sphinx but not by confluencebuilder.I can make my usecase work by adding a suitable `supported_image_types` and adding a call to `self.post_process_images(doctree)` in `prepare_writing()` here: https://github.com/sphinx-contrib/confluencebuilder/blob/01def8c69a51234c5392a69b0022492e950531ed/sphinxcontrib/confluencebuilder/builder.py#L236
(The exact line number is not important, but it needs to run on every documents doctree before the call `self.assets.process(ordered_docnames)`.

@jjcaballero would that solve your usecase? It is much less code than this PR and will keep maintenance of the code that finds the correct file the responsibility of Sphinx, not confluencebuilder.

> I wonder if there is something we could be doing with this to prevent this extension to trying to manage supported image types

@jdknight As Confluence content ends up as HTML anyway, one could assume that the image formats that are supported are as the same for the HTML builder. The list of supported images can then be defined in code with:
```py
from sphinx.builders.html import StandaloneHTMLBuilder 

...

class ConfluenceBuilder(Builder):
    supported_image_types = StandaloneHTMLBuilder.supported_image_types
```

jjcaballero:
@jessetan your approach seems cleaner and works for my use case too.
Do you plan to do a PR for that or should I include it here? 
I would like to try to keep the expected window for this feature mentioned by @jdknight (end of the year 2019)

jessetan:
@jjcaballero if you are willing, it would be great if you could make the changes in this PR. It would save me some work and you have the testcases ready.

After my previous comment, I've found the best place to call `self.env.app.builder.post_process_images(doctree)` is before `processDocument` in https://github.com/sphinx-contrib/confluencebuilder/blob/046ac21f7528b5777826f1c1d21a63bd347f091c/sphinxcontrib/confluencebuilder/assets.py#L138

So far I have found only one downside; builds fail if the PlantUML Sphinx extension is used (https://github.com/sphinx-contrib/plantuml/). The image nodes inserted by that plugin never get a `candidates` attribute.
I've worked around this by adding the following before the call to `post_process_images`
```py
        # Verify every image has a 'candidates' attribute with key '*'
        # This is normally added by Sphinx' ImageCollector, but is not called
        # by sphinxcontrib.plantuml for some reason
        # https://github.com/sphinx-doc/sphinx/blob/089046979f2b84cf850ee95776d10fe9fd9a77e9/sphinx/environment/collectors/asset.py#L43
        for node in doctree.traverse(imageNode):
            if 'candidates' not in node:
                node['candidates'] = { '*': node['uri'] }
```

There must be a better way to do this, but I haven't had time to delve into this issue any deeper. It's probably not necessary to try to fix that in this PR.

jjcaballero:
@jdknight the pull request should be ready now to review/merge. Thanks.

@jessetan eventually, I placed  `self.post_process_images(doctree)`  in  ``prepare_writing() ``  from builder.py. As this method belongs to the parent class sphinx.Builder, IMO it is clearer to call it directly from the child class ConfluenceBuilder, instead of using the self.env.app.builder from the assets.py module.

I agree that the plantuml issue does not need to be addressed here and I hope it does not stop this PR  :-)

jessetan:
@jjcaballero this works for my usecase. Thanks for your effort.

test/unit-tests/common/test_image_star.py
```diff
@@ -0,0 +1,74 @@
...
+"""
+
+from sphinxcontrib.confluencebuilder.state import ConfluenceState
+from sphinxcontrib_confluencebuilder_util import ConfluenceTestUtil as _
+from sphinxcontrib.confluencebuilder.builder import ConfluenceBuilder
+import os
+import re
+import unittest
+import tempfile
+import shutil
```
jdknight: (optional) Re-ordering these imports into lexicographical order would be a nicety.

test/unit-tests/common/test_image_star.py
```diff
@@ -0,0 +1,74 @@
...
+        with the tested extension. Then the document is built.
+        The expected file is created for every image type just replacing
+        a pattern (@REPLACE_EXTENSION@) in one input file
+        '''
+        dataset = os.path.join(self.test_dir, 'dataset-image-star')
+        image_basename = os.path.basename(self.image_file).split('.')[0]
+        for image_type in ConfluenceBuilder.supported_image_types:
+            specifig_dataset = tempfile.mkdtemp()
+            image_extension = re.sub(r'.*/', '', image_type)
+            os.symlink(os.path.join(dataset, 'index.rst'),
```
jdknight: Curious -- will `os.symlink` work in Windows (without additional system platform modifications)?

test/unit-tests/common/test_image_star.py
```diff
@@ -0,0 +1,74 @@
...
+        a temporary sourcesdir is created with the original index.rst
+        and a link to the original image file. The link name will end
+        with the tested extension. Then the document is built.
+        The expected file is created for every image type just replacing
+        a pattern (@REPLACE_EXTENSION@) in one input file
+        '''
+        dataset = os.path.join(self.test_dir, 'dataset-image-star')
+        image_basename = os.path.basename(self.image_file).split('.')[0]
+        for image_type in ConfluenceBuilder.supported_image_types:
+            specifig_dataset = tempfile.mkdtemp()
```
jdknight: `specific_dataset`

test/unit-tests/common/test_image_star.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def write_expected_file(self, pattern_file, outfile, actual_extension):
+        ''' writes the content of pattern_file into outfile replacing
+        the image pattern by the actual extension
+        '''
+        with open(pattern_file) as inp, open(outfile, 'w') as out:
+            content = ''.join(inp.readlines())
+            out.write(content.replace(self.extension_pattern, actual_extension))
+
+    def test_all_supported_extensions(self):
```
jdknight: A suggestion, if I may, for this test. I looks like you are following the convention found in other tests, were you have a data set, generate results and check on an pre-built expected file for verification; however, you are modifying for each image check the expected file and generating new documentation outputs for each type defined in `supported_image_types`. Why not instead just drop `expected-image-star` completely and generate the entire "expected" document and dummy image types in another temporary directory? This could allow you to have just a single dynamic "expected" document which you only need to perform a single Sphinx build event on (rather than building `len(StandaloneHTMLBuilder.supported_image_types)` times).

sphinxcontrib/confluencebuilder/builder.py
```diff
@@ -22,6 +22,7 @@
 from sphinx.errors import ExtensionError
 from sphinx.util import status_iterator
 from sphinx.util.osutil import ensuredir, SEP
+from sphinx.builders.html import StandaloneHTMLBuilder
```
jdknight: (optional) Re-ordering these imports into lexicographical order would be a nicety.

test/unit-tests/common/test_image_star.py
jjcaballero: Good point. I didn't know tests must be cross-platform. I learned os.symlink does not work in Windows for all python versions. Eventually, I am copying the file with shutil module and ensuring the created directory is deleted at the end of the test regardless of the result.

test/unit-tests/common/test_image_star.py
jjcaballero: Addressed.

test/unit-tests/common/test_image_star.py
jjcaballero: Eventually, I realized that to test an image type, you need an actual image with that image type. It seems the Builder discovers the image candidates according to the actual file type, not only the extension. For every candidate, there is a path/uri assigned. I was testing always the same candidate type (png) with different path assigned.

sphinxcontrib/confluencebuilder/builder.py
jjcaballero: Addressed comment. I was not sure if you wanted me to re-order all the imports, I did it, I hope this is ok. I used isort module.

jdknight: Interesting. I was not aware that PEP8 had a recommendation for import ordering. Personally, I am not a fan of this recommendation. I believe (aside from `__future__` requirements) that sorting all imports lexicographically is a much cleaner way to organize imports (for readability); and I never understood why PEP promoted all these extra blank lines.

That being said, maybe I should concede my own personal preferences here. I suspect the majority of Python developers prefer it this way. Also, since Sphinx itself follows this style, I assume it would be good for this extension to follow suit.

test/unit-tests/common/test_image_star.py
```diff
@@ -4,71 +4,88 @@
...
     def setUpClass(cls):
-        cls.extension_pattern = '@REPLACE_EXTENSION@'
-        cls.test_dir = os.path.dirname(os.path.realpath(__file__))
-        cls.image_file = os.path.join(cls.test_dir, 'assets', 'image01.png')
+        test_dir = os.path.dirname(os.path.realpath(__file__))
+        cls.image_file = os.path.join(test_dir, 'assets', 'image01.png')
+        cls.image_basename = cls.file_basename(cls.image_file)
+        cls.datasetdir = tempfile.mkdtemp()
+
+    def tearDown(self):
```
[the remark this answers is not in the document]
jjcaballero: Correct! `tearDown` works too because it is called after every single test. But, as datasetdir is created in `setUpClass`, it makes more sense to clean it in `tearDownClass`.

jdknight:
Go time; thanks all! Merged into `master`; available next stable.

PR #219, opened 2019-07-14
Adding visit node methods - one is called by doxygen directive

jjcaballero:
I have added two visit node methods in translator.py and two unit tests. The code itself is not much, only testing requires a bit more.

**- visit_toctree:** this method is called when there are two nested toctrees and the deepest one is hidden. Other writers (html, html5, manpage) have the same definition. The example below show the structure:

```
# index.rst:

  .. toctee::
     subindex

# subindex.rst

  .. toctree::
    :hidden:

    page
```

**- visit_desc_signature_line:** this method is called by doxygenclass directive when you try to include C++ documentation generated by doxygen. Like:

```
.. doxygenclass::  ClassName
   :project: projectName
   :members:
```

Testing of this requires to have _breathe_ in the extensions of the package (only for testing). I have
 added it to tox.ini for py36 and py37. In any case, I would understand you prefer to keep the environment simpler. I have added it because IMO doxygen is important for many people and being able to have it in this builder can lead to having a global solution for documentation. So, having _breathe_ for testing would make sense for me. I will try to attach a screenshot to show the result of including doxygen documentation in Confluence (from the added unit test) (not sure if this is possible)

I have created a decorators/utils in ConfluenceTestUtil to allow skip unit tests when the python or sphinx version does not satisfy the minim criteria. "skipIf" from pytest skips leads an error.

Thanks for the feedback!


jjcaballero:
doxygen_result

Attached the result of including doxygen documentation on Confluence.
You can compare with the results shown in [breathe documentation](https://breathe.readthedocs.io/en/latest/#in-a-nutshell).

jdknight:
I like a bunch of things in this pull request -- of course the primary fixes to the extension node management and some of the new testing feature checks.

Although, I am personally not a find of the inclusion of `breathe` in the primary test environment. I do not mind having tests for Doxygen (due to its popularity), but the test cases verifying Doxygen output and the use of `breathe` should be done in an independent tox environment. I do not mind also if this environment is included in the Travis CI process, as long as its it an optional test. Support for having Doxygen content placed into documents onto Confluence pages is a nice goal to have, but not a requirement (thus should be optional).

Note that due to other issues being addressed in this extension,  the expected window for this feature is planned around end-of-year (2019; see [this comment](https://github.com/sphinx-contrib/confluencebuilder/pull/215#issuecomment-541268554) for more details. Also note that with possible changes needed for #229, there may need to be additional verification done to make sure any new storage format requirements did not break anything.

jjcaballero:
Thank you, @jdknight.
I understand the point of including breathe and Doxygen in a separate tox environment. As those changes and the changes related to Travis are configuration/management changes, I assume you prefer to do it yourself.

I am really sorry to hear about #229, I hope no new issues have raised since the last update.

jdknight:
Some key changes in this pull request are now being handling in #270. This is to help introduce important changes in this pull request for the upcoming stable release. Reasons why this pull request was not just merged in have been noted above in the comments.

Changes not introduced in #270 should still be considered. The use of breathe will help provide good content to improve visualization of code-related documentation. It maybe more ideal to use this in validation testing instead of unit testing (since there may be too much content for expected output verification to be useful). To be looked at again next development window.

jdknight:
Finally able to complete the port of the remaining features introduced in this pull request (see #478). The validation environment has been tweaked to support processing a series of Doxygen-generated documentation content using Breathe. Hopefully this can help improve autodocs-generated content over the life-cycle of this extension. Thanks again for the original pull request and the fixes/features provided.

PR #221, opened 2019-07-17
 Number of the first line of code can be >1

bosonogi:
Support 'firstline' parameter in the code block macro.

https://confluence.atlassian.com/doc/code-block-macro-139390.html

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,7 @@
 master
 ======
 
+* support 'firstline' parameter in the code block macro
 * fixed issue with document names with path separators for windows users
 * fixed issues when using hierarchy on sphinx 2.1+ (new citations domain)
 * support default alignment in sphinx 2.1+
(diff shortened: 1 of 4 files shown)


bosonogi:
Oops... I'll update the unit tests soon.

bosonogi:
@jdknight Can you take a look at test errors? I'm have no idea why they failed here and passed in #220

jdknight:
> @jdknight Can you take a look at test errors? I'm have no idea why they failed here and passed in #220

A quick glace, it looks like `docutils` has jumped from version 0.14 to 0.15. On examination of `nodes.py`, it seems to import `docutils.utils`, which in turn then imports `from docutils import nodes`. I assume this is why there is an `ImportError`. I will be able to investigate when I get back in August (unless you wish to investigate ahead of time).

jdknight:
Nightly build resolved with the most recent docutils fix; all checked passed. Will merge in.

Thanks!

Issue #136, opened 2018-08-13
Support for math mode


digulla:
The markup page at http://sphinxcontrib-confluencebuilder.readthedocs.io/en/latest/markup.html says that math won't be supported.

Using the plugin LaTeX Math, confluence is able to render math formulas. What is preventing your project from rending the necessary markup?

jdknight:
@digulla, I do not believe there is anything preventing support for reStructuredText's `math` directive in this extension. The directive was listed as "unplanned" due to the amount of other changes to be made (autodoc, images, switch to storage format, etc.).

I see that you have mentioned the [LaTeX Math extension](https://marketplace.atlassian.com/apps/1210882) for Confluence. While I would not be against attempting to add support for Confluence Marketplace extensions, since this is not a stock or freely provided addition to Confluence, other changes need to be considered:
- A user should be able to prevent generating macro content on an instance which does not have the macro available. While we do have a restricted macro option (`confluence_adv_restricted_macros`), I would lean towards a change like this to be more of an opt-in solution.
- There appears to be multiple LaTeX math-based extensions. It may be proper to provide a way to allow a user to configure which LaTeX math-based extension to use.
- Extensions are not free. While I do not mind spending a few coinage here-and-there, trying to add support for these extensions and maintain support for these extensions requires a little finesse.

That all being said, it would be nice to be able to support the `math` directive without needing any additional Confluence extension. Doing a quick scan, I see that there is support in Sphinx (1.8+ native; pre-1.8 via `sphinx.ext.imgmath`) to convert math content into images. With upcoming asset support (#135), it may be possible to support the `math` directive for all users. I would not be opposed to include options which allow a user to use a LaTeX-math supported extension instead of having to build/publish images.

So, possible -- just not a trivial change.

TxBillbr:
Noting a similar problem (maybe the same?) I am getting a sphinx warning running 2.0.0 with confluence builder 1.1.0:
`...\sphinx\transforms\post_transforms\compat.py:54: RemovedInSphinx30Warning: Translator for <class 'sphinxcontrib.confluencebuilder.translator.ConfluenceTranslator'> does not support math_block node'. Please update your extension.
  RemovedInSphinx30Warning)

[...]

PR #244, opened 2019-11-11
Do not fail with NotImplementedError on math_block

jessetan:
Sphinx v2.2+ shows the following warning on every build:
`RemovedInSphinx30Warning: Translator for <class 'sphinxcontrib.confluencebuilder.translator.ConfluenceTranslator'> does not support math_block node'. Please update your extension.`

It does not seem like a problem because [the code that generates this warning](https://github.com/sphinx-doc/sphinx/blob/v2.2.0/sphinx/transforms/post_transforms/compat.py#L50) does not actually check if `math_block` is supported.

However when one does have a document that includes math, the build fails with the following error:
```
Exception occurred:
  File "/usr/local/lib/python3.7/site-packages/sphinxcontrib/confluencebuilder/translator.py", line 132, in unknown_visit
    raise NotImplementedError('unknown node: ' + node.__class__.__name__)
NotImplementedError: unknown node: math_block
```

This PR does not add support for `math_block` (since math is not supported yet, see #136), but it does stop the build from failing by defining a visit function that just skips the node.

### DIFF
--- a/sphinxcontrib/confluencebuilder/translator.py
+++ b/sphinxcontrib/confluencebuilder/translator.py
@@ -1300,6 +1300,10 @@ class ConfluenceTranslator(BaseTranslator):
         # unsupported
         raise nodes.SkipNode
 
+    def visit_math_block(self, node):
+        # unsupported
+        raise nodes.SkipNode
+
     # -------------------------
     # sphinx -- production list
     # -------------------------


jdknight:
@jessetan, thanks for the change. Hopefully we will get proper math-support in the future.

PR #250, opened 2019-11-22
Wrap arguments in option list in <code>

jessetan:
This PR wraps the first table cell of option lists (used to document arguments for CLI programs) in `<code>` so that they are displayed as monospaced text.

Arguments are wrapped in `<em>`. This is in line with how the sphinx_rtd_theme HTML theme does it, but different from stock docutils, which just uses monospace for everything.


jdknight:
Changes look good; I would like to run this through a simple validation check, but I do not believe there will be any issues. Thanks for the tweak.

My only comments about this pull request would be the actual commit messages themselves. This pull request has useful information that would also serve best inside the respective commit messages.

jessetan:
I squashed the commits and reworded the commit message.

jdknight:
@jessetan, thanks for these changed; merged in.

PR #257, opened 2019-11-28
Load custom nodes of autosummary extension

pyracode:
I was trying to use `confluencebuilder` together with the `autosummary` extension. This did not work as the `ConfluenceTranslator` does not know how to handle two custom nodes introduced by `autosummary`: `autosummary_table` and `autosummary_toc`.

This pull request adds this functionality.

Single point of change is the `confluencebuilder/__init__.py`. It checks whether or not `autosummary` is loaded. If so, it adds the missing translation handlers to the nodes.


jdknight:
@pyracode, thanks for the change. I stacked some additional changes (primarily including some validation tests to help with release testing for the `autosummary` directive). Merged into `master` and will be available next stable release.

pyracode:
Thank you @jdknight, glad that I could contribute.

PR #264, opened 2019-12-08
Option to add section numbers to sections and page titles

### LINKED ISSUE #256: Numbered page titles and section headers
This is a poll for interest in the ability to add section numbering to sections and page titles.
I have a rough implementation, but I would like to know if this is within the spirit of this extension before I invest more time.

The use of section numbers is not as applicable to web-based output as Confluence or HTML as it is to printed documents, but it can serve a purpose to indicate a suggested or required reading order.
The main driver for this feature is to add a section number before page titles so that they appear in the Confluence page tree in the correct order. Without this, Confluence will use alphabetical sorting order. Confluence does not support a way to reorder pages using the REST API (#4).
Something similar was started in #164, but this never made it into master.

Possibly related to #195

jessetan:
This PR adds the ability to add numbering to page titles, section headers or both.
To make this work, the `toctree` must have the `:numbered:` option and the new configuration options `confluence_add_secnumbers` and/or `confluence_add_pagesecnumbers` must be enabled.
I've defaulted this to off, so that a `:numbered:` toctree has the same (unnumbered) output as before this PR.

The suffix of the numbering can be changed using `confluence_secnumber_suffix`.

Explicitly excluded in this PR is numbering of figures and tables. This is unchanged.

Fixes #256, and since pages are displayed in lexicographic order in the page tree, also helps #4


sphinxcontrib/confluencebuilder/builder.py
```diff
@@ -323,6 +328,20 @@ def write_doc(self, docname, doctree):
...
+
+        # Add section number to page
+        if self.add_pagesecnumbers:
+            doctitle = self._parse_doctree_title(docname, doctree)
+            if self.secnumbers.get(''):
+                numbers = self.secnumbers['']
+                doctitle = '.'.join(map(str, numbers)) + self.secnumber_suffix + doctitle
+                doctitle = ConfluenceState.registerTitle(docname, doctitle,
+                    self.config.confluence_publish_prefix,
+                    self.config.confluence_publish_postfix)
```
jdknight: I am a bit concerned about this placement of this logic. From my (limited) understanding of the Sphinx engine, a builder's `write_doc` call can be invoked in parallel (especially since we explicitly set `allow_parallel` on the builder). With this logic, you are changing the title name for a `docname` which other documents may have been processed and using older references.

I wonder to address this is as simple as moving this logic into the `prepare_writing` stage; for example:
```
...

doctitle = self._parse_doctree_title(docname, doctree)
if not doctitle:
    continue

if self.add_pagesecnumbers or self.add_secnumbers:
    self.secnumbers = self.env.toc_secnumbers.get(docname, {})

if self.add_pagesecnumbers and self.secnumbers.get(''):
    nums = self.secnumbers['']
    suffix = self.secnumber_suffix
    doctitle = '.'.join(map(str, nums)) + suffix + doctitle

doctitle = ConfluenceState.registerTitle(docname, doctitle,
    self.config.confluence_publish_prefix,
    self.config.confluence_publish_postfix)
    
...
```
This assumes that `self.env.toc_secnumbers` is properly populated at this stage. This change also avoid unnecessarily registered the title twice.

sphinxcontrib/confluencebuilder/translator.py
```diff
@@ -153,10 +155,26 @@ def visit_section(self, node):
...
+            self.body.append('.'.join(map(str, node['secnumber'])) +
+                             self.secnumber_suffix)
+        elif isinstance(node.parent, nodes.section):
+            anchorname = '#' + node.parent['ids'][0]
+            if anchorname not in self.builder.secnumbers:
+                anchorname = ''  # try first heading which has no anchor
+            if self.builder.secnumbers.get(anchorname):
+                numbers = self.builder.secnumbers[anchorname]
+                self.body.append('.'.join(map(str, numbers)) +
+                                 self.secnumber_suffix)
```
jdknight: If we are going to pull code from existing Sphinx implementation, we would need to appropriately add a copyright notice at the top of this file (see also [`confluencebuilder/sphinxcontrib/confluencebuilder/util.py`](https://github.com/sphinx-contrib/confluencebuilder/blob/master/sphinxcontrib/confluencebuilder/util.py#L4) for an example).

sphinxcontrib/confluencebuilder/__init__.py
```diff
@@ -70,6 +70,12 @@ def assetsDocTreeResolvedHook(app, doctree, docname):
     app.add_config_value('confluence_publish_postfix', None, False)
     """Prefix to apply to published pages."""
     app.add_config_value('confluence_publish_prefix', None, False)
+    """Add section numbers if doctree has :numbered: option"""
+    app.add_config_value('confluence_add_secnumbers', False, False)
+    """Add section numbers to page titles if doctree has :numbered: option"""
+    app.add_config_value('confluence_add_pagesecnumbers', False, False)
```
jdknight: I actually wonder if this is best disabled by default. This extension, IMHO, should first and foremost should follow the practices promoted by the core Sphinx implementation and restructuredText specification (when applicable). If a documentation set using `toctree` with the `numbered` option by default generates section identifiers in, let us say, the HTML builder, it would think it would be best for this builder to do the same. While I understand it is not ideal change expected output through various revisions, I do not think the previous limitation of this extension (of not supporting `numbered` should require users properly using the option to also enable the feature as well. If in fact, by default, section numbers are disabled on other builders, this comment can be disregarded.

doc/configuration.rst
```diff
@@ -279,6 +279,37 @@ An example publish prefix is as follows:
 
 By default, no prefix is used. See also |confluence_publish_postfix|_.
 
+confluence_add_secnumbers
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Add section numbers to section titles in pages if ``doctree`` uses the
+``:numbered:`` option. This causes Confluence to display the pages in the correct order in the page tree, since it orders them using lexicographical sorting. By default, this is disabled:
```
jdknight: I am a bit hesitant to promote the idea that using the `confluence_add_secnumbers` option "causes Confluence to display the pages in the correct order". Ordering is a side effect of this feature and there is no guarantee that a publish event will enforce ordering (especially if a node in a page tree is already configured for manual ordering).

If we really want to promote the idea that this option can potentially assist in page ordering, I believe it would be best outlined in a new section under the "tips" documentation.

_Side note:_ it would be ideal if the 80 character column was used for these documents when possible

sphinxcontrib/confluencebuilder/translator.py
jessetan: Good point.

As an alternative, I could directly call `HTML5Translator.add_secnumber()`. That would mean no Sphinx code will need to be copied, but it does tie implementation to the Sphinx implementation (which has upsides and downsides).

Which would you prefer?

sphinxcontrib/confluencebuilder/__init__.py
jessetan: I am of the same opinion, but opted for backward compatibility to increase the chance of this PR being accepted. I changed it to default on.

test/unit-tests/toctree/test_toctree.py
```diff
@@ -38,3 +38,62 @@ def test_toctree_default(self):
...
+        config = dict(self.config)
+        config['confluence_add_secnumbers'] = True
+        dataset = os.path.join(self.test_dir, 'dataset-numbered-depth')
+        expected = os.path.join(self.test_dir, 'expected-numbered-depth')
+        doc_dir, doctree_dir = _.prepareDirectories('toctree-markup-numbered-depth')
+        _.buildSphinx(dataset, doc_dir, doctree_dir, config)
+
+        _.assertExpectedWithOutput(self, 'index', expected, doc_dir)
+        _.assertExpectedWithOutput(self, 'doc1', expected, doc_dir)
+        _.assertExpectedWithOutput(self, 'doc2', expected, doc_dir)
```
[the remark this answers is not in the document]
jessetan: Added

doc/configuration.rst
jessetan: I'm also not sure if this ordering will work the same in all situations, so I removed the text.

sphinxcontrib/confluencebuilder/builder.py
jessetan: I solved it in a slightly different way. I moved the page numbering to `prepare_writing` as you suggested, but I kept the part that called `self.secnumbers = self.env.toc_secnumbers.get(docname, {})` in `write_doc`, as it needs to be done for every doc.
If I didn't, the section numbers available in the writing phase would always be the section numbers of the last processed document.

sphinxcontrib/confluencebuilder/translator.py
jdknight: Copying is fine (with appropriate attribution while also meeting license requirements). Probably better not to rely on other translator's implementation which may cause future issues.

jdknight:
Running this with an `html` builder, a toctree's references will have the section number in the links:

image

However, when running the same content through the Confluence builder, default labelled links on pages will not include section numbers:

image

This expected?

jessetan:
I've changed this so the toctree links now have the correct section numbers. It did involve merging the two options to disable numbering of page titles and section titles into a single option.

sphinxcontrib/confluencebuilder/translator.py
```diff
@@ -1826,6 +1846,16 @@ def _start_ac_link_body(self, node):
         Returns:
             the content
         """
+
+        if self.add_secnumbers and node.get('secnumber'):
+            return  self._start_tag(
+                        node,
+                        'ac:link-body',
+                        suffix=(
```
jessetan: @jdknight Using `suffix` here gives the desired result, even though the section number is technically a prefix. Is this expected?

jdknight: The `suffix` argument is for adding trailing content for a tag being created by a `_start_tag` call (not the content which may be wrapped by tags). Typically, this is for only for specifying newlines for formatting niceties in the output (with the exception of character encoding (`CDATA`) use cases). Real content, like the section numbering, should really be added outside the tag-creation call since the section numbering has nothing to do with the tag itself.

While this change does address adding appropriate prefix content inside generated toctree's, adding this logic into the `_start_ac_link_body` helper call would not be the right place for this. A helper method like `_start_ac_link_body` is much like `_start_tag` which focuses on the "proper" creation of the tag portion for a specific use case, and should not really focus on the content of the data.

I believe a better location for this logic would be inside `_visit_reference_intern_uri`. For example:

```
    def _visit_reference_intern_uri(self, node):
        ...
        self.body.append(self._start_ac_link_body(node))
        if self.add_secnumbers and node.get('secnumber'):
            self.body.append('.'.join(map(str, node['secnumber'])) +
                             self.secnumber_suffix)
        self._reference_context.append(self._end_ac_link_body(node))
        ...
```

jdknight:
@jessetan, I made the one trivial change of moving content from `_start_ac_link_body` to `_visit_reference_intern_uri` (which I outlined above) and rebased the changes on `master` before merging in; I assume that would work for you.

Thanks for the pull request -- nice to have support for numbered sections now. 🎉

PR #311, opened 2020-03-19
Fix images with * extension in singleconfluence builder

jessetan:
If an image uses `*` as an extension, the non-single confluence builder finds a file with the correct extension automatically (#217). This does not work with the singleconfluence builder, it produces the following error:
```
Exception occurred:
  File "<some path>/confluencebuilder/sphinxcontrib/confluencebuilder/util.py", line 34, in hashAsset
    with open(asset, 'rb') as file:
FileNotFoundError: [Errno 2] No such file or directory: '<some path>/image.*' 
```

This PR shuffles some lines to make this work in the singleconfluence builder and adds a test.


jdknight:
Sorry for the long delay here; thanks for the fix.

The whole "preparation" phase is incorrect in the single-builder (which you address here for images by moving `post_process_images`), I will need to move over `_replace_math_blocks` as well.

PR #329, opened 2020-06-12
translator: refactor translator

jdknight:
Refactoring the existing translator to split the implementation between a base translator (`ConfluenceBaseTranslator`) and a new storage format-specific translator (`ConfluenceStorageFormatTranslator`). This is to help prepare for prospect changes where a new translator aims to support the Atlassian Document format.


sphinxcontrib/confluencebuilder/translator/__init__.py
```diff
@@ -0,0 +1,286 @@
+# -*- coding: utf-8 -*-
+"""
+:copyright: Copyright 2016-2020 Sphinx Confluence Builder Contributors (AUTHORS)
+:license: BSD-2-Clause (LICENSE)
+"""
+
+from ..exceptions import ConfluenceError
```
tonybaloney, line 7: can absolute imports be used instead?

jdknight: All relative imports related to this change request have been switched to absolute imports.

Over time, I will try to replace other relative imports in this extension.

Issue #317, opened 2020-04-27
publishing fails for multiple documents referencing same image


g-simmons2:
Hi, I am attempting to publish sphinx documentation to confluence, and encountering this error for pages with images. Pages without images publish with no problems up to the first page with an image, then the following error is thrown and publishing stops. The `"authorized": false` makes me think this could be a permissions error? Any help greatly appreciated.

```
Params
------
looking for now-outdated files... none found
pickling environment... done
checking consistency... done
preparing documents... done
writing output... [100%] validation/validation                                                                                  
publishing documents... [  7%] overview/overview                                                                                
sphinxcontrib.confluencebuilder error:
---
An unsupported Confluence API call has been made.

REQ: PUT
RSP: 500
URL: https://company.atlassian.net/wiki/rest/api
API: content
DATA: {
  "statusCode": 500,
  "data": {
    "authorized": false,
    "valid": true,
    "errors": [],
    "successful": false
  },
  "message": "com.atlassian.confluence.api.service.exceptions.InternalServerException: java.lang.RuntimeException: com.atlassian.confluence.content.render.xhtml.storage.link.StorageLinkUnmarshalXhtmlException: java.lang.UnsupportedOperationException"
}
---
```

g-simmons2:
Found out this only occurs if multiple rst files are referencing images of the same filename, so solved by copying and changing the filename of images that I would like to include in multiple locations.

jdknight:
@g-simmons2, thanks for the report; will try to find a correction.

ottowayi:
For me, just changing the file name did not work.  I had a bunch of SVGs that were reused (like button icons).  I made a unique copy for each place one was used, then opened them in a text editor and added some extra whitespace in a different spot for each image.  They still render the same, but I think making them hash uniquely solved the problem.

g-simmons2:
@ottowayi iirc I noticed something similar in certain cases, where changing the filename didn't solve the problem and I had to change the file contents. I also recall looking into the `confluence_asset_override` configuration option, can't remember whether it was successful or not. 

https://sphinxcontrib-confluencebuilder.readthedocs.io/en/latest/configuration.html#confluence-asset-override

ottowayi:
@g-simmons2 I tried setting `confluence_asset_override = True` and it didn't help.  I only got it to work once I made sure the file contents were unique, not just the file names.

jdknight:
Found it.

[...]

PR #334, opened 2020-07-26
storage: do not inject newline inside attachment tag

jdknight:
It has been observed that (in at least) Confluence Cloud that when a newline is injected in the `ri:attachment` tag, typically only when `ri:content-title` is used, that Confluence will report a 500 error when publishing. This commit removes the suffix injection (from an explicit newline to an implicit empty/none) and publish events appear to work as expected.

See also #317.


claco:
Is there a new release of 1.2.1 pending? This is currently hitting us. Thank you!

jdknight:
@claco, the current plan is to have a v1.3 release between sometime this November and December. Unfortunately, a specific date is not set at this time. The only workaround I can suggest at this time is to [install this extension's from the most recent development state](https://sphinxcontrib-confluencebuilder.readthedocs.io/en/stable/install.html#master).

Issue #316, opened 2020-04-17
Clean prefixes


adityamj:
Hi,

I am trying to have clean prefixes as demonstrated at https://jdknight.atlassian.net/wiki/spaces/confluencebuilder/overview

The demonstration has clean top level prefix like 1.2v and other pages have 1.2v-

I am unable to remove the hyphen from the top level index page. My top level page looks like "1.1v-"

Is there a way to have clean prefixes like you have shown in the demonstration?

Thanks,
Aditya

jdknight:
@adityamj, the [`confluence_publish_prefix`](https://sphinxcontrib-confluencebuilder.readthedocs.io/en/latest/configuration.html#confluence-publish-prefix) option should be just a direct prefix on document title values, so a configuration such as:

```
confluence_publish_prefix = '1.2v-'
```

Should be a translated to `1.2v-<generated-title-value>`. Based on your comments, I would suspect that either the following is happening:

1. You may have [`confluence_add_secnumbers`](https://sphinxcontrib-confluencebuilder.readthedocs.io/en/latest/configuration.html#confluence-add-secnumbers) enabled (by default) and which may not actually be a desired feature for your use case.
2. You may have invoked a publish attempt multiple times with various `confluence_publish_prefix` values; however, this will not remove older published documents by default (see also [`confluence_purge`](https://sphinxcontrib-confluencebuilder.readthedocs.io/en/latest/configuration.html#confluence-purge)).

If neither is what you are experiencing, could you then provide more information? Hopefully we can sort out the root cause here (ambiguous configuration/documentation or an actual bug in the extension).

adityamj:
Hi, thanks for taking the time to go through this issue! 

So here is my setup: 

1. Each tag/branch gets its own prefix. For example master branch has
the prefix 'master-', v3.1 has prefix 'v3.1-' and so on.
2. All these versions are uploaded under a page in confluence. 

Here is how the menu structure looks: 

* Top level page
   * master-
      *master-introduction
      *master-some other page etc
   * v3.1-
      *v-3.1-introduction 

and so on. 

Secnumbers are disabled in my toctree so that is not the issue. 

What I would like is for the menu to look like: 

* Top level page
   * master
      *master-introduction
      *master-some other page etc
   * v3.1
      *v-3.1-introduction 

Here, the "-" from prefix is not there in branches' top level page
however it is still there in pages lower in the tree. 

This is just an aesthetic issue so that top level pages don't have a
[...]

PR #349, opened 2020-10-26
introduce disabling prefix/postfix on index page

jdknight:
A user may wish to opt-out of adding a prefix or postfix to an index page if the title is dynamic for their publish event. This commit introduces the configuration entry `confluence_ignore_titlefix_on_index` to provide this flexibility.


jdknight:
See also #316.

Issue #353, opened 2020-11-02
Broken link at `confluence_ca_cert`


danieltuzes:
No page for [Requests SSL Cert Verification](http://docs.python-requests.org/en/master/user/advanced/#ssl-cert-verification) is found in the [documentation at `confluence_ca_cert`](https://sphinxcontrib-confluencebuilder.readthedocs.io/en/stable/configuration.html#confluence-ca-cert)

jdknight:
The Requests documentation appears to be updated to be working now on "python-requests.org". The documentation has also been updated to use the "requests.readthedocs.io" domain (see #359). It is assumed that the latter is "preferred" since it appeared to be the most recent update (and has a "stable" release point).

In any case, I believe this issue has been resolved. Thanks for the report.

PR #359, opened 2020-11-26
Documentation: update requests-related documentation links

jdknight:
Requests documentation is now found under the "requests.readthedocs.io" domain (instead of under a "*.python-requests.org" domain). Adjusting to the new domains and point to the stable variant of the respective documentation sections.


jdknight:
See also #353.

Issue #346, opened 2020-10-20
Links to sections with section number does not work


pawelkami:
I'm using ```sphinxcontrib-confluencebuilder==1.2.0``` from pip.

Confluence builder does not properly publish links to sections of the document when they use section numbers.
It works properly for documents without section numbers.

For example, we have a similar RST document:

```
Some installation manual
----------------------------

This is some installation manual.

Follow these steps to install the software:

#. `Install AAA`_
#. `Configure BBB`_
#. `Install CCC`_


Install AAA
~~~~~~~~~~

aaaaaaaaaaaaaaaaaaaaaa

Configure BBB
~~~~~~~~~~~~~

bbbbbbbbbbbbbbbbbbbbbbbb

Install CCC
~~~~~~~~~~~~

cccccccccccccccccccccc

```

When this is built it looks like that:

```
9. Some installation manual

This is some installation manual.

Follow these steps to install the software:

1. Install AAA (hypertext)
2. Configure BBB (hypertext)
3. Install CCC (hypertext)


9.1. Install AAA

aaaaaaaaaaaaaaaaaaaaaa

9.2. Configure BBB

bbbbbbbbbbbbbbbbbbbbbbbb

9.3. Install CCC

cccccccccccccccccccccc

```

Hypertext links to these subsections look this way:
```http://some.wiki.address/display/namespace/9.+Some+installation+manual#id-9.Someinstallationmanual-InstallAAA```, 
but it doesn't work for confluence, because it expects this link 
```http://some.wiki.address/display/namespace/9.+Some+installation+manual#id-9.Someinstallationmanual-9.1.InstallAAA```.

When I turn off section numbers (```confluence_add_secnumbers = False```), links are generated in the same way, but they work.
Your plugin is missing section number support for links.

jdknight:
The changes outlined in #365 are now available on the main branch and will be available next stable release (v1.3+). Will mark as pending-next-release and will close up when the release has been completed (see #362 for more details).

jdknight:
v1.3 is now available on PyPI -- marking as closed.

PR #365, opened 2020-11-28
improve support for section/figure numbering

jdknight:
The following provides a more complete for section numbering support when used with this extension. This extension's builders never built proper targets for sections to reference against. Section names identifiers are internally managed by the Confluence instance and the target mappings never took into consideration of a section with a prefixed section number. Cached target entries will now handle when a section number is added to a section text.

In addition, the `singleconfluence` builder required implementing its own title cache list, as the section targets need to be compiled after assembling the section numbers, but before the primary doctree is built. This is a result of pre-merged doctree's which will point to page-specific targets (e.g. `sub/#target`) but the original implementation would assume only an index-based target existed (e.g. `index/#target`) -- which could result in conflicts.

In addition, Sphinx's figure numbering capability was partially ported over when section number was added. This commit attempts to complete the use of the feature by ensuring the main builder registers a fignumbers instance for the translator and add prefixes on figures (in the storage format) when configured  with the common [`numfig` option](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-numfig).

### DIFF
--- a/sphinxcontrib/confluencebuilder/builder.py
+++ b/sphinxcontrib/confluencebuilder/builder.py
@@ -63,13 +63,10 @@ class ConfluenceBuilder(Builder):
     def __init__(self, app):
         super(ConfluenceBuilder, self).__init__(app)
 
-        # section numbers for headings in the currently visited document
-        self.secnumbers = {}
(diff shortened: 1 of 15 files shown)


jdknight:
See also #346.

Issue #281, opened 2020-02-01
sphinx.ext.todo does not generate proper links


jdknight:
When using `sphinx.ext.todo` extension, references generated in a todo list do no properly point to the respective todo entries.

jdknight:
With changes introduced from #369, links should be properly created when using the `sphinx.ext.todo` extension. Changes are available on the main branch and will be available next stable release (v1.3+).

PR #369, opened 2020-11-29
storage: handle anchor-specific targets flagged as internal

jdknight:
In select cases, it has been observed that a reference node may contain a `refuri` definition which contains an anchor to the active document being processed. The existing implementation would only be able to handle this case if the reference was also not marked as internal; however, it has been observed that in select cases (such as using the `sphinx.ext.todo` extension) that local anchor entries can also be on references which have an internal flag set.

This commit re-works the process `refuri` process flow by:

1) Always checking for an local anchor reference to forward to to the internal identifier-based reference implementation (the fix).
2) If not internal, continue to forward it to the external reference     implementation.
3) Otherwise, process it as an internal URI.


jdknight:
See also #281.

Issue #304, opened 2020-02-26
NotImplementedError: unknown node: autosummary_table


engineerjoe440:
Ran into an issue upon my first test-run with this module. From other issues, it seems like this has already been managed, so maybe I'm just missing something. I'm still pretty sphinx illiterate, so any help would be greatly appreciated!

Full Traceback:

```
# Sphinx version: 2.3.1
# Python version: 3.8.1 (CPython)
# Docutils version: 0.16 release
# Jinja2 version: 2.10.3
# Last messages:
#   
#   looking for now-outdated files...
#   none found
#   pickling environment...
#   done
#   checking consistency...
#   done
#   preparing documents...
#   done
#   writing output... [ 14%] docs/AcRTAC
# Loaded extensions:
#   sphinx.ext.mathjax (2.3.1) from c:\users\joestan\appdata\local\programs\python\python38-32\lib\site-packages\sphinx\ext\mathjax.py
#   sphinxcontrib.applehelp (1.0.1) from c:\users\joestan\appdata\local\programs\python\python38-32\lib\site-packages\sphinxcontrib\applehelp\__init__.py
#   sphinxcontrib.devhelp (1.0.1) from c:\users\joestan\appdata\local\programs\python\python38-32\lib\site-packages\sphinxcontrib\devhelp\__init__.py
#   sphinxcontrib.htmlhelp (1.0.2) from c:\users\joestan\appdata\local\programs\python\python38-32\lib\site-packages\sphinxcontrib\htmlhelp\__init__.py
#   sphinxcontrib.serializinghtml (1.1.3) from c:\users\joestan\appdata\local\programs\python\python38-32\lib\site-packages\sphinxcontrib\serializinghtml\__init__.py
#   sphinxcontrib.qthelp (1.0.2) from c:\users\joestan\appdata\local\programs\python\python38-32\lib\site-packages\sphinxcontrib\qthelp\__init__.py
#   alabaster (0.7.12) from c:\users\joestan\appdata\local\programs\python\python38-32\lib\site-packages\alabaster\__init__.py
#   sphinx.ext.autodoc (2.3.1) from c:\users\joestan\appdata\local\programs\python\python38-32\lib\site-packages\sphinx\ext\autodoc\__init__.py
#   sphinx.ext.napoleon (2.3.1) from c:\users\joestan\appdata\local\programs\python\python38-32\lib\site-packages\sphinx\ext\napoleon\__init__.py
[...]

PR #370, opened 2020-11-29
inject autosummary nodes after a proper extension check

jdknight:
This extension would attempt to register compatible autosummary nodes when the `sphinx.ext.autosummary` extension was loaded. This would allow a user to take advantage of the autosummary capabilities with this extension.

The problem with the check on `app.config.extensions` is that that configuration file only indicates extensions the user wishes to explicitly load, and not what extensions have been actually loaded into the Sphinx engine.

The injection process has been changed to better handle autosummary support:

1) Explicitly check against the `app.extensions` loaded extensions to determine if the nodes should be injected.
2) Only attempt to inject after the `builder-inited` event to ensure that modules loaded after the Confluence builder extension which may load `sphinx.ext.autosummary` extension themselves does not cause an missed injection.

### DIFF
--- a/sphinxcontrib/confluencebuilder/__init__.py
+++ b/sphinxcontrib/confluencebuilder/__init__.py
@@ -216,8 +216,13 @@ def setup(app):
     """Wires up the directives themselves"""
     app.add_directive('confluence_metadata', ConfluenceMetadataDirective)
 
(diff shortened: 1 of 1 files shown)


jdknight:
See also #304.

Issue #318, opened 2020-05-11
unknown node: graphviz


RiccardoManfrin:
This is similar to #304 

`
Exception occurred:
  File "/usr/local/lib/python3.8/site-packages/sphinxcontrib/confluencebuilder/translator.py", line 149, in unknown_visit
    raise NotImplementedError('unknown node: ' + node_name)
NotImplementedError: unknown node: graphviz
`

adrianohrl:
I've tried to use the `sphinx.ext.graphviz` along with the `sphinxcontrib.confluencebuilder`; although, a `NotImplementedError` is prompted to me. The same as reported above.

Is there any update to this issue since its submission?

adrianohrl:
As a temporary solution for this, I've attached the following snippet in the project `conf.py` file:

```python
from graphviz import Source, render


def convert() -> None:
    suffix = '.svg'
    path = Path('_static/graphs').resolve()
    for path in path.rglob('*.dot'):
        print(f'Reading the {path} file ...')
        graph = Source.from_file(path)
        print(f'Writing the {path}{suffix} file ...')
        render(path.suffix[1:], suffix[1:], path)


convert()
```

It converts the all `DOT` file in the `_static/graphs` folder to `SVG`. Then, while documenting, I can use the `image` and `figure` directives pointing to the converted files. 

> Note: You may include `**/*.dot.svg` to the project's `.gitignore` file.

jdknight:
The changes outlined in #373 are now available on the main branch and will be available next stable release (v1.3+). Will mark as pending-next-release and will close up when the release has been completed (see #362 for more details).

jdknight:
v1.3 is now available on PyPI -- marking as closed.

PR #373, opened 2020-12-01
initial support for graphviz graphs

jdknight:
This commit introduces the initial support for graphviz graphs. The implementation mimics the same implementation used to support inheritance diagrams \[1\], where graphviz nodes are converted into image nodes before being processed by a translator.

With the example documentation:

```
.. graphviz::

   digraph foo {
      "bar" -> "baz";
   }

.. graph:: foo

   "bar" -- "baz";

.. digraph:: foo

   "bar" -> "baz" -> "quux";
```

A final output renders as follows:

> image

\[1\]: 01f97a67b98d598a6f5a542869f8542d08fd931b

### DIFF
--- a/doc/markup.rst
+++ b/doc/markup.rst
@@ -82,6 +82,7 @@ The following extensions are supported:
 
  - `sphinx.ext.autodoc`_
  - `sphinx.ext.autosummary`_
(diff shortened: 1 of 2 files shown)


jdknight:
See also #318.

Issue #338, opened 2020-08-05
Support as target of intersphinx


Holi0317:
Seems that current version supports linking out to other sphinx page with [sphinx.ext.intersphinx]. But does this plugin supports letting other pages to link to a sphinx-built confluence page?

[sphinx.ext.intersphinx]: https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html

jdknight:
@Holi0317, it may be safe to say this extension does not properly work with the `sphinx.ext.intersphinx` extension. I cannot say I have used the intersphinx extension myself, but maybe it would be possible to generate a compatible `objects.inv` file on a full process event (although, I only have done a quick pass at the front-end documentation without knowing the full effort involved).

One possible limitation that might exist is the inability to support the `intersphinx_mapping` configuration, since this extension is limited to adding files as attachments on pages -- so there may not be a graceful way to link to an `objects.inv` file.

jessetan:
The `intersphinx_mapping` option can take a location of `objects.inv` and a different root url for the documentation pages itself. So theoretically, this extension could generate an `objects.inv` file which can be stored locally or saved to some central HTTP server manually.

The syntax for `objects.inv` is relatively straightforward: https://sphobjinv.readthedocs.io/en/latest/syntax.html

From Sphinx documentation:
> A dictionary mapping unique identifiers to a tuple (target, inventory). Each target is the base URI of a foreign Sphinx documentation set and can be a local path or an HTTP URI. The inventory indicates where the inventory file can be found: it can be None (at the same location as the base URI) or another local or HTTP URI.
https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html#confval-intersphinx_mapping

jdknight:
A goal has been added to include a limited version of this for the v1.3 release. The plan is to help generate a series of `std:doc` and `std:ref` entries for documents (pages) and title entries for the initial release. From here, we can expand on this capability for more advanced cases. An initial implementation is planned in the upcoming days.

jdknight:
The changes outlined in #377, e45870bae6722967dc0176cbfcbafcc5dd75fdc5 and 8d13f305d4055ab201f0bcbf4cd1ed0c84f27214 are available on the main branch and will be available next stable release (v1.3+). Will mark as pending-next-release and will close up when the release has been completed (see #362 for more details).

bskinn:
[...]

PR #377, opened 2020-12-05
initial intersphinx support

jdknight:
The following adds initial support for generating an inventory file to be used by the `sphinx.ext.intersphinx` extension. This initial release will build document and various references (e.g. titles) inside an inventory file.

This file is only built when publishing occurs, as it is the only time it will know the specific page identifiers to build links for targets.


jdknight:
See #338.

Issue #410, opened 2020-12-14
Error when a .rst file contains only an empty section


adreyfus:
Hi,

Thank you for your excellent work.

I encountered a problem with the .rst files which only contain a section title, without content. 

For example :

index.rst :

```
.. _configuration_index:

Configuration
#############

```
The file index.conf is not generated, then the error  

`WARNING: error reading file C:\.../config/index.conf: [Errno 2] No such file or directory: 'C:\\.../config/index.conf'
` 

appears, and sections are missing in the result sent on Confluence.

To get around it, of course, you just have to put something in the section, like "to be completed". But I report this because I guess this is not voluntary behavior :) .

jdknight:
Thanks for reporting this. This should now be supported on the main branch (via #411) and should be made available next stable release (v1.3).

jdknight:
v1.3 is now available on PyPI -- marking as closed.

PR #411, opened 2020-12-14
builder: write empty pages

jdknight:
In the event where a document entry only contains a title reference and is removed via the `confluence_remove_title` option, the resulting page will be empty. While this should be fine, the existing builder implementation would not generate an output document which was empty. This would leave to an error such as the following:

    WARNING: error reading file<filename>.conf: [Errno 2] No such file or directory: ...

If a documentation set has an empty page, this extension should still be creating/publishing it. Correcting this in the builder to write content if even an empty string is set.


jdknight:
See also #410.

Issue #448, opened 2021-03-08
:doc: reference incorrectly cross-referenced when using singleconfluence builder


tsvi:
I have a multiple-page document, but I need to publish the documentation as a single page.
When using the `:doc:` directive for referencing, the singleconfluence builder points the anchor to the filename instead of the section name.

See the following demo:

Source file tree:
 - index.rst
 - overview.rst - has a pointer to example.rst
   ```
   See the section on :doc:`Writing an XML <example>`
   ```
 - example.rst

Output conf file:
```
see the section on <ac:link ac:anchor="example">
<ac:link-body>Writing an XML</ac:link-body>
</ac:link>
```

When looking at the link generated by confluence:
- redacted/display/Area/PageName#PageNameTitle-**example**
But the anchor points to:
- redacted/pages/viewpage.action?pageId=1789830485#PageNameTitle-**Writinganxml**

jdknight:
@tsvi, thanks for reporting this issue. A change (#454) has been added into the main branch that should address this issue and should be made available next stable release (v1.5).

If you still experience this issue after this change, please feel free to let know.

jdknight:
v1.5 is now available on PyPI -- marking as closed.

PR #454, opened 2021-04-10
ensure doc refs in singleconfluence work

jdknight:
When building a single Confluence page, if a reference explicitly references to another document (`:doc:<>`), there can be no registered target to properly point to. The `_register_doctree_title_targets` call focuses on building targets based off the identifiers defined in the section names; however, there is no guarantee that the assigned identifiers will match to an internal refid value to a specific document name (e.g. if "pagea" links to "pageb", page A can have a reference to a refid "pageb", but page B's root section name *may* not have a matching identifier).

Since we want references like this to point to the starting location of these document's content, there is a desire to register a reference to the first section name value (i.e. first title) if one exists. Therefore, if the root section name does not register a target which matches a prospect refid value for a `:doc:<>` reference, register an additional target to the leading section which has this mapping.

In addition, when `confluence_remove_title` is set (default), while an additional target can be registered for the top section, none will exist for references to the root document if the title is removed. In the event that this scenario is detected, replace any document references of the root pages with a `#top` link instead.


jdknight:
See also: #448

Issue #445, opened 2021-03-02
Support for Markdown Tables


arwedus:
In general, confluencebuilder exports Markdown input which has been parsed by the `recommonmark` or the `myst_parser` extension just fine. Unfortunately, tables are not supported (recommonmark would support either plain HTML tables or the pipetables via `sphinx_markdown_tables`).

I understand that those extensions are not part of sphinx.ext and therefore not officially supported, but it seems to be possible to support, as all that recommonmark or myst_parser do is map to docUtils format in the end, and all other Markdown features I've used so far with both parsers are supported.

The following is an example for a Markdown table, showing which combinations of input and output work for tables:

|              | reStructuredText | recommonmark (md) | MyST (md) |
|--------------|:----------------:|:-----------------:|:---------:|
| HTML         |   yes            |  yes              |   yes     |
| Confluence   |   yes            |  no               |   no      |

jdknight:
_(note: the following was only looked at using recommonmark)_

Honestly, this extension (at least, from a development aspect) has never really looked at support with respect to Sphinx's existing support with Markdown. Any capabilities working inside the `confluencebuilder` extension now works by chance.

Taking a look at the table support via the `sphinx_markdown_tables` extension, it appears the process is as follows:
1. Hook onto a `source-read` event to look for detected tables (via a Markdown table extension).
2. Replace the contents of the Markdown table block with raw HTML.
3. Replace any processed source with the new one before running the source through a builder/translator.

This works for Sphinx's `html`/`singlehtml` builders since it looks like Sphinx will detect the HTML content in the Markdown file and wrap it in a `raw` node type flagged as `html` content. When `html`/`singlehtml` translators process these nodes, it will just dump them in the generated documents.

The reason why these tables are ignored by the `confluencebuilder` extension is that when using storage format translator (i.e. when attempting to translate a document set into Confluence storage format compatible output), the raw type identifier is not set to a `confluence_storage` value. Supported storage format markup is not completely compatible with HTML/HTML5/etc. markup, and this extension ensures the raw types do not mix by explicitly requiring Confluence-marked raw type identifiers to be set (which allows users to customize unique output between HTML and confluence builders, if desired).

There can be an agreement that it is annoying since the generated output partially works on a Confluence instance if the raw HTML types are permitted (with some implementation tweaks). One could hack a way to permit the inclusion of Markdown tables by using the following inside a `conf.py` file:

```
from docutils import nodes
[...]

PR #475, opened 2021-05-29
support cell alignment hints injected in the classes attribute

jdknight:
It has been observed that some extensions (e.g. MyST-Parser) will inject alignment hints for cells into a node's `classes` attribute. This would allow a styling to define a, for example, `text-align:center` class which can include a center alignment hint for formatting on a document.

In the event that a text-align style has been detected inside a class list on a node, directly inject it into the style attribute of the paragraph holding the cell data.


jdknight:
(see also #445)

PR #504, opened 2021-10-16
ci: add black and isort

mschoenlaub:
I noticed that there are no pe-commit hooks (and therefor no obvious way to see how the project is linted) and formatting is a bit awkard and I decided to introduce black and isort, as they are both de-facto standard tools and they provide the ability to actually fix deviations that they find.

Would you be interested in that?


mschoenlaub:
@jdknight What's your opinion on this?

jdknight:
Hey @mschoenlaub, thanks for the pull request. There are a chunk of things I like from these changes, but also a chunk of things I am not a fan of. I will comment and ask more specific questions below as I nitpick through this entire change set.

---------------

First off, I do not fully understand your statement here:

> there are no [pre]-commit hooks (and therefor no obvious way to see how the project is linted) 

This project uses flake8 (as I am sure you have seen based off your changes) and is invoked by default via a `tox` call as promoted in the [CONTRIBUTING](https://github.com/sphinx-contrib/confluencebuilder/blob/master/CONTRIBUTING.rst) document. Are you suggesting here that this project should define a [pre-commit framework](https://pre-commit.com/) configuration that developers could use? If that is the suggestion, I am indifferent on such an inclusion; as long as it is not a burden on the maintenance of this project. Personally, I do not think I would use it myself, but if it helps others who may wish to develop on this extension, that works for me.

> formatting is a bit awkard

I would agree that the subjectiveness of preferred code styling can be awkward (for any project/language). This project is not an exception -- while the use of PEP8 is promoted, a [series of recommendations are ignored](https://github.com/sphinx-contrib/confluencebuilder/blob/ef93320ead496a4e55458dd46a2a23669f62f17a/setup.cfg#L21-L32) due to preferences. I can acknowledge that this list have been mostly tailored to my own preferences (being the active maintainer). There are some styling choices which I can be more flexible on, if they are more preferred by the community. For example, I was initially not a fan of introducing [two spaces for inline comments](https://www.python.org/dev/peps/pep-0008/#inline-comments) and I am still not a not a fan of [grouping imports into sections](https://www.python.org/dev/peps/pep-0008/#imports), but I have no objections if others wanted to enforce these. However, several linting recommendations I am cannot promote. One, for example, I am adamant on never grouping imports on the same line when the imports come from the same module. I am also not a big fan of various newline recommendations, such as injecting two lines between things like functions, etc. or enforcing newlines in code blocks which could (subjectively) be better visualized/readable in a more compact way (although I will relax this preference based on some discussion below).

Let us dive into the actual changes.

---

I have never been a big fan of automated code formatting. A lot of my bias from this spawned from my previous experiences using code formatters in some C/C++ projects a long time ago. However, I would not object an attempt to introduce it in this project. It appears that some changes [Black](https://black.readthedocs.io/) performs (subjectively) make the implementation look nicer. That being said, I cannot say I prefer the switch from single-quotes to double-quotes. I would recommend if there is a desire to introduce the Black code formatter to this project, its invoke should include the `--skip-string-normalization` argument for now. The debate on a preferred quoting style could be discussed in another issue and not block its introduced into the development branch.

There are a bunch of changes which I like. My favourite changes are related to the formatting of various function/method calls. For example, the following is much cleaner (opinion):

```diff
-            for asset in status_iterator(assets, 'publishing assets... ',
-                    length=len(assets), verbosity=self._verbose,
-                    stringify_func=to_asset_name):
+            for asset in status_iterator(
+                assets,
+                'publishing assets... ',
+                length=len(assets),
+                verbosity=self._verbose,
+                stringify_func=to_asset_name,
+            ):
```

The individual lines makes it clearer and I love the introduction of a trailing comma (for future line-diff minimization). I kinda wished Black would also lexicographically order the keyword arguments as well (but that is most likely outside the context of this merge request).

I am not a big fan of the injection of various newlines (as mentioned above). For example:

```diff
         class MockTranslator:
             def __init__(self, builder):
                 self.builder = builder
+
         mock_translator = MockTranslator(self)
```

Or:

```diff
 try:
     from sphinx.util import progress_message
 except ImportError:
+
     class progress_message:
         def __init__(self, msg):
             self.msg = msg
```

However, I know a lot of Python projects really use/like this spacing -- so maybe its best that I concede to such changes.

---

I would like to suggest that Black be configured to respect an 80-character line-limit (`--line-length 80`). I know this project disables this check for flake8 (`E501`); however, this is due to not having a gracefuly way to invoke flake8 to ignore some long line lengths which are preferred to be longer (such as long imports, more about this below; without getting into adding various noqa comments in the source). If this option could be set via `pyproject.toml`, that would be great -- assuming it does not cause issues for the existing options set in `setup.cfg` (if so, maybe any migrations/changes here would be best looked at in another merge request).

---

I cannot say I am a fan on how Black formats some slices. For example, this looks like a good change to me:

```diff
-            self.docparent = self.docname[0:self.docname.rfind(SEP) + 1]
+            self.docparent = self.docname[0 : self.docname.rfind(SEP) + 1]
```

However, this looks like an odd change to me:

```diff
-                refnode['refid'] = refuri[idx2 + 1:]
-                del refnode['refuri']
-            elif refuri.startswith(root_docuri + '#'):
-                refnode['refid'] = refuri[idx + 1:]
-                del refnode['refuri']
+                refnode["refid"] = refuri[idx2 + 1 :]
+                del refnode["refuri"]
+            elif refuri.startswith(root_docuri + "#"):
+                refnode["refid"] = refuri[idx + 1 :]
+                del refnode["refuri"]
```

Since it only happens in this block of code, I guess it is not a big deal (but I guess this is an example of why I am not a fan of code formatters sometimes).

---

As I have mentioned above, I do not agree with the grouping of imports on the same line when the imports come from the same module. I believe it best to leave individual imports on their own individual lines. It does not seem possible adjust this with Black (intended by the design of the code formatter). If Black were to be introduced, it would be nice to have isort to be worked in a way to support a more extensive lint check. The following configuration change would be the recommended configuration (in `setup.cfg`) for the isort tool:

```
[isort]
force_single_line = 1
line_length = 999
```

This would promote a more community desired section grouping of imports (which I do not mind changing to), but still preventing the grouping of module-specific imports and multilining these imports.

A question I have is, is it possible to use such a configuration and ensure a Black-invoked linter check could still pass? (e.g. run an isort check to ensure imports are "right", re-run isort to re-format the imports in a way Black would like and then finally running a Black check to ensure the implementation is in a "good" state).

Would it be ideal to introduce a `format` (or whatever may be the common convention) environment to tox to allow developers to invoke all code-formatter changes with a single environment call (e.g. `tox -e format`)?

---

Lastly, I do not mind introducing a general `lint` environment into the tox configuration and adjusting the GitHub actions we have to invoke it. Invoking just a single linter job instead of multiple per linter can be fine; I do not know how much is gained by splitting them up (note though, in the proposed changes, it would be nice if the `toxenv` entries were aligned with each other).

It would be nice in `tox.ini` that the existing `pylint` environment could still remain. This can provide flexibility in developers wishing to only invoke `pylint` to do so.

It does not appear that in the introduced `lint` environment that `flake8` is invoked at all. Also, it looks like `E203` would need to be configured in `setup.cfg` to [prevent flake8 from complaining about format changes made by Black](https://black.readthedocs.io/en/stable/faq.html#why-are-flake8-s-e203-and-w503-violated).

---

Feel free to comment/critique.

jdknight:
Closing this merge request.

Is it the opinion of the current active maintainer (myself) that the resulting formatting changes by Black are not desired at this time (this may change in the future on how this project evolves). Note that the this merge request was appreciated though -- several styling/formatting changes have been made to cleanup the implementation, and some originally suppressed linter checks have been restored/enforced.

Issue #443, opened 2021-02-26
Links using labels or :contents: doesn't work


samuelabreu:
Hi, im using sphinxcontrib-confluencebuilder==1.4.0

When i create an index on a file using:
```
.. contents::
   :local:

```

or internal links using refs:

```
.. _my-label:

Click here :ref:`my-label`
```

or even links to sections as:

```
My Section
-------------

Go to `My Section`_
```

links are created, works on html, but on confluence all those links go anywhere, checking the url it changes to:

`https://confluence/display/SPACE/Pagename+with+space#MySection`

Reading https://support.atlassian.com/confluence-cloud/docs/insert-links-and-anchors/ it should be

`https://confluence/display/SPACE/Pagename+with+space#Pagenamewithspace-MySection`

Couldn't find if its a config missing on sphinx or confluence itself, any hint how to solve that?

gec37:
Hi @jdknight 

Thanks for looking into this.

The issue of additional white space when using `.. contents ::` seems to become apparent only if you have a long string for the title of the page and/or the section headers.

The title always seems to be included in the table of contents whether you use the `:local:` directive or not.

Although the links render as expected when using `:ref:` (with the sphinx.ext.autosectionlabel extension enabled), if you click on any of the links, the first time seems to direct you to the correct part of the page, but after that the links appear broken (i.e. do not re-direct you to the correct location on the page when you click on them).

jdknight:
By chance, does the root `toctree` directive use the `:numbered:` option?

I do see an issue with references when this option is set.

jdknight:
An update on this -- if the reported issues are related to using `:numbered:` in a toctree directive, this issue should be resolved with the changes introduced in #545,

If this is not the case, I have unfortunately not been able to reproduce this issue. As mentioned in some of the comments above, attempts have been made to use [title entries with spaces or extremely large names](https://raw.githubusercontent.com/sphinx-contrib/confluencebuilder/70d41cd01dbebd03c9cbb191d72d75619e4925a8/tests/unit-tests/datasets/toctree-numbered/doc.rst). At this time, the inspected output shows expected results.

If anyone has any example documentation which produces this issue (with relevant configuration options), please let me know. I will in turn be able to locally produce it and add some unit tests to help cover this case.

---

A side note (if I may), if a document is of limited size, for example:

```
simple
======

.. contents::
    :local:


section
-------

this is a test
```

And the total page of a Confluence document is viewable in the view box of a browser:

[...]

PR #545, opened 2021-11-11
use proper identifier for section number mapping

jdknight:
Section numbers populated into the environment's `toc_secnumbers` dictionary is based off the section identifier value -- not the text content of a section (used for mapping to Confluence's "target"). When attempting to determine which section number value to apply for a target, use the proper identifier instead.


jdknight:
#443

PR #621, opened 2022-02-20
test: active test server asap

jdknight:
When preparing a dummy HTTP test server for mocked server validation, a quick client may observe connection resets if the server fails to activate before requests are issued. To help prevent this, explicitly active the server (when not ignoring requests) before creating a thread to serve requests.

---

This change is in response to observed GitHub action failures reporting (e.g. [run 5266167550](https://github.com/sphinx-contrib/confluencebuilder/runs/5266167550)):

```
ConnectionResetError: [Errno 54] Connection reset by peer
```

It is assumed that requests are being issues before sockets are completely activated (to accept connections); although, this is an assumption. It appears to only occur on Windows/OSX containers at this time, so there could be another issue being overlooked.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,33 +30,8 @@ jobs:
     strategy:
       matrix:
         include:
-            - { os:  ubuntu-latest, python:  "2.7", toxenv:  py27-sphinx18, cache: ~/.cache/pip }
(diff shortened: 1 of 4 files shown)


jdknight:
Dropping ~ after additional inspection of the test logic, the changes presented in the pull request were incorrect.

PR #654, opened 2022-05-25
Fix for going over CONFLUENCE_MAX_TITLE_LEN while handling title conflict

AntonMihalkov:
Increasing trimming length over CONFLUENCE_MAX_TITLE_LEN results in failed publishing attempt. Instead, trimming using tail length so that the title won't go over the limit. Signed-off-by: Anton Mihalkov <box4android@gmail.com>

### DIFF
--- a/sphinxcontrib/confluencebuilder/state.py
+++ b/sphinxcontrib/confluencebuilder/state.py
@@ -107,9 +107,8 @@ class ConfluenceState:
                         ConfluenceState.title2doc[title.lower()], docname))
 
             tail = ' ({}){}'.format(offset, base_tail)
-            try_max = CONFLUENCE_MAX_TITLE_LEN + len(tail)
-            if len(base_title) > try_max:
-                base_title = base_title[0:try_max]
+            if len(base_title) + len(tail) > try_max:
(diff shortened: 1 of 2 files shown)


AntonMihalkov:
Added a signed off message in my commit, yet there still seems to be a problem according to the DCO. Please advise.

jdknight:
Thanks for the pull request. For sure the original implementation was just broken. I had one concern with respect to this change. For these lines in the original implementation:

https://github.com/sphinx-contrib/confluencebuilder/blob/99f2658504fcad0f22c1dd139e1c822e01cffa38/sphinxcontrib/confluencebuilder/state.py#L91-L93

I assume when a post-fix is set, it can result in the same problem with bumping `try_max` (I assume the original implementation should have been subtracting this length instead). If `try_max` is improperly set, I assume it would still cause issues with the corrects you are introducing in the title-conflict check.

Any chance you are willing to look at that? I could look at trying to add some unit tests later to sanity check title trimming/duplication logic (unless you'd be interested in adding some as well).

AntonMihalkov:
@jdknight don't plan to add any unit tests at this point

jdknight:
@AntonMihalkov, thanks for the pull request. 🙇

PR #673, opened 2022-07-03
Feature: make headers and footers be real templates

rvodden:
I'd like to be able to include the version number of the project in a footer. This PR creates two new configuration options `confluence_header_data` and `confluence_footer_data`. When each of these are set, then the contents of `confluence_header_file` and `confluence_footer_file` is interpreted to be a jinja template. A typical usage might be:

contents of `assets/header.tpl`
```
version: {{ version }}
```

contents of `conf.py`:
```python
// ...

confluence_footer_file = 'assets/header.tpl'
confluence_footer_data = {
    version = "1.2.3"
}
// ...
```
It may well be possible to use the metadata builtin module to load the version number dynamically.


sphinxcontrib/confluencebuilder/builder.py
```diff
@@ -12,6 +12,7 @@
 from docutils import nodes
 from docutils.io import StringOutput
 from os import path
+from jinja2 import Environment, BaseLoader
```
jdknight, line 15: A potential recommendation, if I may, is that maybe we do not need to explicitly import/reference the jinja2 engine. Sphinx already provides [some glue](https://github.com/sphinx-doc/sphinx/blob/master/sphinx/jinja2glue.py) to utilize jinja2 in Sphinx, and Confluence builder [should be binding a jinja2 environment/base loader automatically](https://github.com/sphinx-contrib/confluencebuilder/blob/master/sphinxcontrib/confluencebuilder/builder.py#L133-L134). An example case we have in this extension is to [utilize template processing](https://github.com/sphinx-contrib/confluencebuilder/blob/fe503c0f4619f80febfd0a0c362ba5c45c5c41c3/sphinxcontrib/confluencebuilder/storage/index.py#L52-L58) for a configured index page.

Maybe, instead of:

```
rtemplate = Environment(loader=BaseLoader).from_string(template_data)
self._cached_header_data = rtemplate.render(**self.config.confluence_header_data)
```

Something like this would work (note: not tested this):

```
output = self.templates.render_string(template_data.decode('utf-8'),
    self.config.confluence_header_data)
```

If we can do this, it would could:

- Drop the need for these new imports.
- Drop the need to add the `jinja2` package added to `requirements_validation.txt`.

sphinxcontrib/confluencebuilder/builder.py
```diff
@@ -885,11 +886,18 @@ def _generate_special_document(self, docname, generator):
...
                     with io.open(fname, encoding='utf-8') as file:
-                        self._cached_header_data = file.read() + '\n'
+                        template_data = file.read() + '\n'
                 except (IOError, OSError) as err:
                     self.warn('error reading file {}: {}'.format(fname, err))
 
-        # cache footer data (if any)
+                # if no data is supplied, the file is plain text
+                if self.config.confluence_header_data is None:
+                    self._cached_header_data = template_data
```
jdknight, line 895: There is a potential of `template_data` to be undefined here if the file read attempt results in an exception.

Maybe configure `template_data` to an empty string instead of `self._cached_header_data` at the start of these scopes; or switch it back to just set `self._cached_header_data` and overwrite the value for the template case (either way works for me).

Same would apply for the footer processing section.

sphinxcontrib/confluencebuilder/translator.py
```diff
@@ -95,10 +96,19 @@ def depart_document(self, node):
...
             try:
                 with io.open(header_file, encoding='utf-8') as file:
-                    self.body_final += file.read() + self.nl
+                    header_template_data = file.read()
             except (IOError, OSError) as err:
                 self.warn('error reading file {}: {}'.format(header_file, err))
 
+            # if no data is supplied, the file is plain text
+            if self.builder.config.confluence_header_data is None:
+                header = header_template_data
```
jdknight, line 105: (same idea as other comment) `header_template_data` may be undefined.

Also applies to `footer_template_data`.

sphinxcontrib/confluencebuilder/translator.py
```diff
@@ -7,6 +7,7 @@
 from docutils import nodes
 from docutils.nodes import NodeVisitor as BaseTranslator
 from os import path
+from jinja2 import BaseLoader, Environment
```
jdknight, line 10: (same idea as other comment) Switch to embedded template engine.

tests/unit-tests/templates/sample-footer-with-jinja.tpl
```diff
@@ -0,0 +1 @@
+footer content {{ variable }}
```
jdknight, line 1: (optional)

In these examples and test code, can these be updated to include the new line at the end of these files.

doc/configuration.rst
```diff
@@ -247,7 +247,16 @@ Generic configuration
 
         confluence_header_file = 'assets/header.tpl'
 
-    See also |confluence_footer_file|_.
+    See also |confluence_footer_file|_, |confluence_header_data|_.
+
+.. confval:: confluence_header_data
+
```
jdknight, line 253: (optional) Add a `.. versionadded:: 1.9` entry.

doc/configuration.rst
```diff
@@ -263,6 +272,15 @@ Generic configuration
 
         confluence_footer_file = 'assets/footer.tpl'
 
+    See also |confluence_header_file|_, |confluence_footer_data|_.
+
+.. confval:: confluence_footer_data
+
```
jdknight, line 278: (optional) Add a `.. versionadded:: 1.9` entry.

jdknight (changes requested):
This is a great idea. And still promotes compatible with possible older template configurations.

Added a series of recommendations, if you are willing. If these adjustments are made (or there are some reasons that the original change is better), I do not see any reason why this cannot be merged in. Feel free to comment/critique these recommendations.

rvodden:
An excellent set of suggestions. I'll incorporate all these and resubmit :-)

doc/configuration.rst
```diff
@@ -247,7 +247,16 @@ Generic configuration
 
         confluence_header_file = 'assets/header.tpl'
 
-    See also |confluence_footer_file|_.
+    See also |confluence_footer_file|_, |confluence_header_data|_.
```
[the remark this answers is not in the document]
rvodden: I've done this :smile:

sphinxcontrib/confluencebuilder/builder.py
rvodden: I've done this too, the decode didn't seem to be necessary, and in the translator I had to grab the templates object from the builder, but it is certainly much cleaner without that extra dependency!

sphinxcontrib/confluencebuilder/builder.py
rvodden: Done

sphinxcontrib/confluencebuilder/translator.py
rvodden: Done

sphinxcontrib/confluencebuilder/translator.py
rvodden: Done

tests/unit-tests/templates/sample-footer-with-jinja.tpl
rvodden: I have done this, Can I suggest that you perhaps include a `.editorconfig`? (See here: https://editorconfig.org/) That would allow you to explicitly stipulate the need for new lines :-)

rvodden:
@jdknight All yours. Thank you very much for the incredibly fast review!

rvodden:
Needs a rebase before review.

rvodden:
@jdknight All yours :-) Thank you very much for the incredibly fast review!

sphinxcontrib/confluencebuilder/translator.py
```diff
@@ -91,14 +91,25 @@ def depart_document(self, node):
 
         # prepend header (if any)
         if self.builder.config.confluence_header_file is not None:
+            header = ''
```
jdknight, line 94: I assume this was suppose to be `header_template_data`?

You'd also want to make sure the same changes would apply to `footer_template_data`.

jdknight (changes requested):

sphinxcontrib/confluencebuilder/translator.py
rvodden: yup - fixed in both cases. Thanks!

rvodden:
Yay formatting issues - 1 moment :-)

rvodden:
Should be fixed now @jdknight

jdknight:
@rvodden, thanks for the changes; much appreciated.

Issue #676, opened 2022-07-07
Allow multiple pages with the same title


finkedave:
Confluence does not allow this, which is noted in this repos documentation. This project allows prefixes and suffixes to help alleviate a bit of the pain when trying to push generated pages into spaces that have manually generated pages with the postfix and prefix configs. 

The issue arrises is what if I want multiple pages with the same title throughout my source directory that I am trying to push to confluence. ie readme pages for different directories, how-tos,  results for analyses etc.  People suggest using manual suffixes for each page that has the same title but that is too much of pain for us. So the question is how to do this within this project but still make the suffixes / prefixes consistent even different people run sphinx and push to confluence. You wouldn't want it to keep purging pages and creating new pages and thus destroying links if different people run it.

My answer was create a hash to the path to the source file and append it to the end of the title when registering the title for a page. So if readme is in different subfolders it will be a different hash and its title will be something like readme -xk3432zt and readme -zpo778d34 .   I also have a path relative part of this since different user's are going to have a different absolute path.

I am currently finishing my testing of this feature but wanted to know if others had a better solution and/or if this is something that other's might want in the main fork.

jdknight:
If in a situation where multiple projects are processed and published to a single Confluence space (either through a "main" Sphinx project with sub-projects; or multiple independent Sphinx projects), I can agree that it can be less than ideal in trying to prevent/manage conflicts. If individual projects do not wish to manage their own prefix/postfix entries, conflicts are bound to arise for common document titles (as already mentioned).

Aside from having projects managing prefix/postfix overrides for a project (even for versions of a project), the only other pattern I have seen is projects naming their documents with unique/nesting patterns. For example, the leading titles named something such as:

     libfoo/About
     libfoo/Releases/1/ChangeSet
     libfoo/Releases/2/ChangeSet
     libfoo/Releases/3/ChangeSet
     libfoo/SomeCapability

But this is most likely not ideal for most projects, especially when mixing both Confluence publications with another format.

Your prospect work into generating a hash for a document to postfix is not a bad idea. It is a similar pattern used in environments where account identifiers are postfixed with trailing hash/code. The only gain I see off hand is that instead of each project assigning a custom postfix value in their configuration, they would instead apply a flag (or, in your prospect case, a base-path) in their configuration without needing to care about what the postfix value is. There would be no guarantee that conflicts will not occur, but depending on the hash length, maybe it could be rare.

[...]

PR #680, opened 2022-07-14
Allowing multiple pages with the same title

finkedave:
Adding an option to confluence_publish_postfix for making it dynamic and thus allowing users to be able to publish pages with the same titles in sub directories

Resolved ticket: #676 

@jdknight Thanks for looking at the ticket and giving the concept your thoughts. Totally good ideas on how to actually calculate the hash. Since the key piece is that it has to be the same regardless of who runs it.

I am going to open this PR to show how I originally have it working on our system. That whole confluence_title_hash_root_path PR that i closed was.a total mess and never meant to open a PR with it. This way is so much cleaner since i just expand on your usage of postfix

But again I am open to all of your suggestions for changing this if there is a possibility of getting this feature into the base repos.


jdknight (changes requested):
Thanks for the pull request. I do like this idea of utilizing the existing postfix implementation to help apply a generated hash value. I have some recommendations on this approach, if you are willing.

The proposed changes would have a project define the following to utilize hash generation for titles:

```
confluence_publish_postfix = '&unique_hash'
```

Which could generate documents such as `My Title -SF6F5AG`. This has two limitations:

- A user does not have the flexibility to configure how many hash characters to accept.
- Cannot format the trailing hash value to another pattern (e.g. `My Title (SF6F5AG)`).

I wonder if it would be better if we could make this more powerful by allowing a user to format the postfix to decide if they wish to accept a hash value. Consider the following snippet:

```
hash='da39a3ee5e6b4b0d3255bfef95601890afd80709'[:10]
fmt = "My postfix"
print(fmt.format(hash=hash)) # outputs: My postfix
fmt = "My postfix ({hash})"
print(fmt.format(hash=hash)) # outputs: My postfix (da39a3ee5e)
fmt = "My postfix ({hash:.5})"
print(fmt.format(hash=hash)) # outputs: My postfix (da39a)
```

And maybe in the future, additional options can be added into the format string if a user wished to take advantage of them.

I have also adding a series of specific change requests/comments, if you are willing.

jdknight:
In addition, I forgot to mention that the original commit was not signed with a DCO. I would recommend [reading how to address this](https://github.com/sphinx-contrib/confluencebuilder/pull/680/checks?check_run_id=7349336683), since it is a requirement for changes to be pulled into this extension.

jdknight:
Note that there may be a possibility of shipping out a v1.9 release of this extension next weekend (2022-07-[30-31]). If this is a feature you would like in the upcoming release, please let me know and we can try to ensure we can review/fix things up to get this merged in. Note that it is important that the changes are rebased with the DCO set, or I can neither merge this in or try to cleanup any of the proposed changes (if you wanted to have a maintainer finish the work). Otherwise, we can wait for the next following release to try to get this feature in.

finkedave:
> Note that there may be a possibility of shipping out a v1.9 release of this extension next weekend (2022-07-[30-31]). If this is a feature you would like in the upcoming release, please let me know and we can try to ensure we can review/fix things up to get this merged in. Note that it is important that the changes are rebased with the DCO set, or I can neither merge this in or try to cleanup any of the proposed changes (if you wanted to have a maintainer finish the work). Otherwise, we can wait for the next following release to try to get this feature in.

Sorry was on vacation. I wasn't able to find the time to make the changes yet, going to try to spend some time this week on them. All your suggestions are spot on, I just need to find the time to do it.   No rush on getting this into the next release, but I will make sure I have all this done by next week some time so I can get it onto the following release.

finkedave:
I changed the unit testing and methods a bit to make it easier for others to add additional placeholders. Great comments on making this more robust.

sphinxcontrib/confluencebuilder/state.py
```diff
@@ -210,3 +220,32 @@ def upload_id(docname):
...
         """
         return ConfluenceState.doc2uploadId.get(docname)
+
+    @staticmethod
+    def _format_postfix(postfix, docname, config):
+        """
+        Format a postfix that may have placeholders.
+        All placeholders used must be supported otherwise an error is raised
+        """
+        if postfix and postfix_string_replacement.match(postfix):
```
jdknight, line 230: Out of curiosity, does the use of regular expressions here help that much? I assume it was added to help avoid attempts to invoke a format operation (e.g. generating hashes) if no keywords were detected. I would have been happy if we just performed a format request on a non-empty string (even if not keywords were provided), but if the use of `re` helps here, works for me.

That being said, this implementation is calling `match` per title being processed. I would imagine if you are looking for performance here, we could change this to check the postfix once for any possible keywords. And if any are detected, set a state/class flag to be check on each `_format_postfix` call (also, ensuring the flag is cleared in the `reset()` call). Although, I am not worried about it. If you only want to use the implementation as is, that is fine by me.

Lastly, I believe `search` may be a faster alternative to `match` for the small string literals a postfix value could be (based on some casual browsing on the two types).

doc/configuration.rst
```diff
@@ -616,6 +616,19 @@ Publishing configuration
...
+
+    Currently supported placeholders
+
+    * {hash} - Create a reproduceable hash given the title and location based 
+               from the project root. Using this placeholder provides an option
+               for allowing pages with the same title to be pushed to the same 
+               confluence space without needing to manually add an index 
+               to the title
+
+    Note, pages with the same title must have different parent pages.
```
jdknight, line 630: > [Pages] with the same title must have different parent pages.

I do not believe this note is applicable anymore (correct me if I am wrong here).

doc/configuration.rst
```diff
@@ -616,6 +616,19 @@ Publishing configuration
 
        confluence_publish_postfix = '-postfix'
 
+    Postfixes can include placeholders. These placeholders are filled using the
+    format method so formatting types can be used.
+
+    Currently supported placeholders
+
+    * {hash} - Create a reproduceable hash given the title and location based
```
jdknight, line 624: While this text is nicely formatted for the raw content, this will not format as expected in the rendered documentation. The trailing lines for the bullet point need to be aligned with the first line's leading content. In addition, it may be nice to keyword the `{hash}` value (if desired). For example:

```
    * ``{hash}`` - Create a reproduceable hash given the title and location based
      from the project root. Using this placeholder provides an option
      ...
```

doc/configuration.rst
```diff
@@ -616,6 +616,19 @@ Publishing configuration
 
        confluence_publish_postfix = '-postfix'
 
+    Postfixes can include placeholders. These placeholders are filled using the
+    format method so formatting types can be used.
+
+    Currently supported placeholders
```
jdknight, line 622: Maybe end this with a colon? Or just:

```
    Supported placeholders:
```

doc/configuration.rst
```diff
@@ -616,6 +616,19 @@ Publishing configuration
...
+
+    * {hash} - Create a reproduceable hash given the title and location based 
+               from the project root. Using this placeholder provides an option
+               for allowing pages with the same title to be pushed to the same 
+               confluence space without needing to manually add an index
```
jdknight, line 627: `confluence` -> `Confluence`

tests/unit-tests/datasets/postfix_formatting/index.rst
```diff
@@ -0,0 +1,8 @@
+:orphan:
+
+Index page
+==========
+
+.. toctree::
+    readme
```
jdknight, line 7: I assume there should be a newline here between the directive definition and its contents.

tests/unit-tests/datasets/postfix_formatting/index.rst
```diff
@@ -0,0 +1,8 @@
+:orphan:
```
jdknight, line 1: Since this is an index page with a toctree, there is no need to add an `:orphan:` metadata (the same for the child documents that are explicitly referenced inside the `toctree` directive.

tests/unit-tests/test_config_postfix_formatting.py
```diff
@@ -0,0 +1,119 @@
...
+"""
+
+import os
+from sphinxcontrib.confluencebuilder.exceptions import \
+    ConfluenceConfigurationError
+from sphinxcontrib.confluencebuilder.state import ConfluenceState
+from tests.lib.testcase import ConfluenceTestCase
+from tests.lib import parse
+from tests.lib.testcase import setup_builder
+from unittest.mock import patch
```
jdknight, line 14: Note that we still support Python 2.7, so the mock module will not be available.

jdknight (changes requested):
Thanks for the updates, the changes are looking good to far. Going to stack on some additional comments based off the new change set.

I'll also trigger the runners.

sphinxcontrib/confluencebuilder/state.py
finkedave: I'm good with keeping this file simple. But ya my reasoning is if we keep adding more replaces placeholders I didn't want to make people using static suffixes builds continually slower. I'll remove all the regex stuff to keep the change straightforward then in the future if more placeholders are added we always could add regex back in and only calculate things that have placeholders present. ie don't calculate hash unless {hash} was found in the postfix. So it would be a dynamically called format

tests/unit-tests/test_config_postfix_formatting.py
finkedave: I added the requirement for mock on python < 3.3. To keep the tests clean. If you would perfer not to use mock for this repos due to the python <3.3 requirement I can alter these tests as well to just use a known hash for the docname its not a big deal for me either way

jdknight: 👍, I have no problem introducing a `mock` requirement for older Python versions for the test environment.

jdknight (approved):

PR #681, opened 2022-07-15
Recognize confluence_publish as a boolean config

tucked:
This corrects the behavior of `-D confluence_publish=false`.


jdknight:
The pull request is appreciated; good find on the missing entry.

While this was planned to be merged in (after the DCO was addressed), I had some time to look at the environment-related configuration support for this extension. With the changes introduced in #682, decided to cleanup how configuration options were initialized/tracked to prevent missing options like this again. Now, when configurations are registered in this extension, bool/int types will be tracked and will be properly managed via command line processing (avoiding the risks of missing configuration options as they get added/changed over time).

Since there is no longer a list of tracked options in the "defaults" implementation, will mark this pull request as closed.

Issue #679, opened 2022-07-14
Configuration via Environment Variables


tucked:
We have found ourselves frequently adding plumbing for plugin configs in `conf.py` e.g.
``` python
def getenvbool(*args: Any, **kwargs: Any) -> bool:
    """Read an environment variable as a boolean."""
    return os.getenv(*args, **kwargs).lower() in [
        "true",
        "1",
    ]


confluence_publish = getenvbool("confluence_publish", "false")
confluence_publish_dryrun = getenvbool("confluence_publish_dryrun", "false")
confluence_purge = getenvbool("confluence_purge", "false")

confluence_server_user = os.getenv("confluence_server_user")
confluence_server_pass = os.getenv("confluence_server_pass")
confluence_ask_user = confluence_server_user is None
confluence_ask_password = confluence_server_pass is None

...
```
After not working on the project for quite awhile, I returned to it and wanted to set `confluence_publish_postfix`, but I had forgotten about this plumbing. To get it to work, I had to add
```  python
confluence_publish_postfix = os.getenv("confluence_publish_postfix", "")
```
This got me thinking that it might just be handier to generalize this plumbing:
``` python
for key, value in os.environ.items():
    if key.startswith("confluence_"):
        if value.lower() in {"true", "false"}:
            value = bool(value)
        locals()[key] = value
```
However, that's not perfect since some configurations accept `int`s, `list`s, or `dict`s...

Would this project consider adding some help for this?
Perhaps it could accept any config via the environment or offer a function that reads config from the environment and returns it as a properly-typed dictionary e.g.
``` python
# no fancy type conversion needed!
locals().update(confluencebuilder_env_config())
```

jdknight:
Typically in Sphinx environments, configuration overrides from outside `conf.py` are set [using the `-D` command line argument](https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-D). However, I do not see a reason why we cannot make things even more convenient for users by supporting options from the environment as well.

To support something like this, I would imagine it would involve the following:
- Processing and assigning configuration options before the [configuration check](https://github.com/sphinx-contrib/confluencebuilder/blob/a7e2ffdb5d6e06a3e198a8353211cd629f2f014d/sphinxcontrib/confluencebuilder/builder.py#L94).
[...]

PR #682, opened 2022-07-17
support providing configuration options from environment

jdknight:
Provides a series of changed to help support accepting configuration options from the environment.

### track configuration options for configuration type conversions

A subset of options to be used as a boolean or integer type can be accepted as a string (typically, from command line) and converted after post-processing configuration options. The original implementation defined a list of, for example, boolean configuration options and cycled through these keys to ensured all string-assigned values are converted to boolean types.

A maintenance issue observed is that as boolean options are defined for the extension, development may miss adding these new options into the respective type conversion lists in the "defaults" implementation. To prevent this from happening, rework this extension's configuration processing to help easily track the expected types for individual configurations. With this change, the "defaults" implementation can apply type conversions based off tracked configuration entries registered in the manager.

### accept configuration options from environment

If a Confluence supported configuration option is not configured, allow accepting an option from the running environment. Provides flexibility for users wishing to configure a subset of options using environment variables, instead of passing options via the command line or trying to manipulate a project's `conf.py` file to read options from the environment.

### tests: sanity check support for configuration from environment

Adding a series of tests verifying the ability to accept configuration options from the environment.

### doc: document options from environment

Providing information on how users can provide configuration options using the system environment.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -21,6 +21,15 @@ All configurations introduced by this extension are listed below. This
 extension may take advantage of a subset of `Sphinx configurations`_ as well
 when preparing documents.
 
+.. versionadded:: 1.9
+
+    All options provided by this extension may be set from the running
+    environment. For example, if ``confluence_publish`` is not explicitly set
+    inside ``conf.py`` or provided via `Sphinx's command line`_, this extension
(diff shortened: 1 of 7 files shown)


jdknight:
See also #679.

Issue #678, opened 2022-07-12
Use `.netrc` on Unix systems for authentication


alexonea:
Hello maintainers of `confluencebuilder`! 

I am happily using your software for a while and I started thinking about the possibility to allow the plugin to read credentials from the `.netrc` file on Unix systems.

I browsed a bit the sources and I believe that one way to integrate the new functionality would be here
https://github.com/sphinx-contrib/confluencebuilder/blob/c075c91e5756b2d9c2ad99a9b3d1d3954d63ce57/sphinxcontrib/confluencebuilder/rest.py#L214-L220

In case a user is provided through `confluence_server_user` and `passwd is None`, then instead of assuming `passwd = ''` we can attempt to get the password from `.netrc`.

Of course, in principle this might break some code/usage for people who have empty passwords (maybe). In practice, I believe that this is not a problem since someone having an empty password for a certain username/host combination, for sure won't have a `.netrc` entry for the same user/host with a password also specified.

Looking forward to your thoughts!

jdknight:
@alexonea, since this extension uses the Requests library, [support for `.netrc`](https://requests.readthedocs.io/en/latest/user/authentication/#netrc-authentication) should already work without additional configurations. Specifically, by omitting the inclusion of the `confluence_server_user` and `confluence_server_pass` variables, no session auth instance is created and Requests can default to a user's netrc file (if any). For example, if trying to publish to the following URL:

```
confluence_server_url = 'https://myawesomesite.atlassian.net/wiki/'
```

One can authenticate to this instance with a `.netrc` file looking like the following:

```
machine myawesomesite.atlassian.com
login myuser@example.org
password APITOKEN
```

Is there a specific case where you are seeing where this is not working (that I may be overlooking)? I assume at the very least, this extension's documentation could possibly hint at netrc support alongside the authentication options.

PR #684, opened 2022-07-17
doc: provide information about netrc support

jdknight:
Requests supports netrc authentication, but a user may not know exactly how to utilize it with various authentication options for this extension. Add an indication that `confluence_server_user` can be omitted, which would allow Requests to fallback to netrc (if other advanced authentication options are not set).


jdknight:
See also #678.

PR #737, opened 2023-01-01
support collapse hints for literal blocks

jdknight:
If a processing literal block builds a code macro, adding support to flag the `collapse` parameter option if the `collapse` class is specified.

---

fyi @senges, best to create new issue entries for new enhancements than update a closed issue when possible. The features/changes you proposed look good, and have been included in this pull request. This should be available next release. Please feel free to comment if there is anything missing.

### DIFF
--- a/doc/features.rst
+++ b/doc/features.rst
@@ -102,9 +102,6 @@ Type                    Notes
 `versionchanged`_       Supported
 ======================= =====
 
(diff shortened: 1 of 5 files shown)


senges:
Glad to hear that, thanks !

Tbh I never thought this would be included one day in confluencebuilder, that's why I left is as "a tip" on another issue.
I will make sure to create a new Issue next time then.

Issue #761, opened 2023-01-27
confluence_version_comment does not work on initial publish


tucked:
While using the plugin to develop content, I published to my personal space (for the first time), and I noticed that the comment in `confluence_version_comment` (in this case, the Git commit) was not present. I re-published the same content, though, and it was!
![image](https://user-images.githubusercontent.com/48501491/215199052-45fb6c62-da02-4454-b3e8-a3cc1560f257.png)

jdknight:
For sure the existing implementation [only injects a comment for updates pages](https://github.com/sphinx-contrib/confluencebuilder/blob/27547021b019994d70957476d7a681bc905a42d9/sphinxcontrib/confluencebuilder/publisher.py#L977). I can take a look to see if the API supports setting a message value for new pages -- it was either missed, which we can fix; or Confluence ignores the option (which we should then update documentation to reflect the limitation). Will investigate.

tucked:
It _looks_ like `_update_page` should be getting called for new pages:
https://github.com/sphinx-contrib/confluencebuilder/blob/27547021b019994d70957476d7a681bc905a42d9/sphinxcontrib/confluencebuilder/publisher.py#L739-L741
I think the issue is that the new page does not overwrite `page`:
https://github.com/sphinx-contrib/confluencebuilder/blob/27547021b019994d70957476d7a681bc905a42d9/sphinxcontrib/confluencebuilder/publisher.py#L688

jdknight:
> new page does not overwrite `page`

Note that this is done purposely in the implementation. The `_update_page` is solely for processing a page update. The reason for the `if page:` check is to deal with (1) an initial detected page update and (2) detecting a page update from a non-case matching title after failing to upload an original-detected new page.

----

Did some testing with Cloud and server installs. It looks, on a Cloud installation, if we do populate a `version` entry for a new page entry, with a version of `1` with a message field, the page history will be populated with the provided message. Unfortunately, it does not look to be the case for non-Cloud installations. We can queue up a fix for Cloud installations, and add some notes in the configuration documentation to indicate that limitation for non-Cloud instances (until a fix or workaround is found).

jdknight:
Note that I have completed the support request with Atlassian about the injection of a comment for new pages on Confluence Server/DC variants. I have been informed of the following:

> [Confluence Server/DC does not] allow a version comment for the initial page creation in the web UI, or via the REST API.
> _(reference: CA-2218680)_

[...]

PR #762, opened 2023-01-28
publisher: ensure new pages include a configured version comment

jdknight:
When a user configures `confluence_version_comment`, the value used will only show in a page's history for page updates. This is due to the original implementation only populate the version message for page updates. This commit adjusts the implementation to always populate a `version` entry for a page with a configured message value.

Note that new page comments only appear to work on Confluence Cloud. Tests with Confluence server (v7.19.x) does not appear to accept a message value (ignored?).

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -1557,6 +1557,14 @@ Advanced publishing configuration
 .. confval:: confluence_version_comment
 
     .. versionadded:: 1.8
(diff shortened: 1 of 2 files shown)


jdknight:
Will perform some additional tests to see if we can get the initial message working in Confluence server. If not, another commit should be stacked on documenting the limitation.

See also #761.

PR #793, opened 2023-04-18
Initial idea to handle Confluence Links

### LINKED ISSUE #790: Are smart links supported?
Hello folks! I've been using this amazing extension and I've been wondering if Smart Links on editor v2 are supported.
Been trying for a while with regular markdown links (Using myst parser) and with the raw role and the link storage format, but I couldn't manage to turn the link into a Smart Link right away.

Thanks a lot (=

IceS2:
My initial idea on how we could start handling Confluence Smart Links. My intention is to start a discussion around this and maybe collaborate with you (=

It's just a small start since it is not supporting all the possibilities...

One one side, I've thought about having one role for each `data-card-appearance` and deal with the text similar to what Ref does.

Also there's  the possibility to actually use `<ac:link>` tags for the Links still in the editor v2.... I'd imagine we could expand it with that somehow...

Last but not least, I'm not sure this works for editor v1.

Thanks a lot o/

closes #790

### DIFF
--- a/doc/directives.rst
+++ b/doc/directives.rst
@@ -532,6 +532,136 @@ Confluence page.
 
 See also :ref:`LaTeX roles <latex-roles>`.
 
+.. index:: Smart links (directive)
+.. _smart-link-directives:
+
+Smart links
+-----------
+
+.. note::
+
+    Smart links will only render when using the v2 editor
+    (see ``confluence_editor``; :ref:`ref<confluence_editor>`).
+
+.. rst:directive:: confluence_doc
+
+    .. versionadded:: 2.1
+
+    The ``confluence_doc`` directive allows a user to define a link to a
+    document that is styled with a card appearance. The directive accepts the
+    name of a document in an absolute or relative fashion (in the same manner
+    as Sphinx's `:doc: <role-doc_>`_ role). For example:
+
+    .. code-block:: rst
+
+        .. confluence_doc:: my-other-page
+
+    This directive supports the following options:
+
+    .. rst:directive:option:: card: card type
+        :type: block, embed
+
+        When using ``block``, a smart link card is generated for the document
+        link. The card will provide a summary of the target document. When
+        using ``embed``, the contents of the document will rendered on the
+        page.
+
+        .. code-block:: rst
+
+            .. confluence_doc:: my-other-page
+                :card: block
+
+    .. rst:directive:option:: layout: layout type
+        :type: align-start, align-end, center, wrap-left, wrap-right
+
+        .. note:: This option is only applicable when using an ``embed`` card.
+
+        Specifies how an embedded card will be laid out on a page. Embedded
+        cards will default to 100% width. Therefore, to take advantage of
+        certain layout capabilities, users should also assign an appropriate
+        width as well.
+
+        .. code-block:: rst
+
+            .. confluence_doc:: my-other-page
+                :card: embed
+                :layout: align-end
+                :width: 20
+
+    .. rst:directive:option:: width: value
+        :type: integer
+
+        .. note:: This option is only applicable when using an ``embed`` card.
+
+        Specifies the width to apply for an embedded card. The width is a value
+        from 0 to 100 (e.g. a value of ``80`` for 80% of the page).
+
+        .. code-block:: rst
+
+            .. confluence_doc:: my-other-page
+                :card: embed
+                :width: 50
+
+.. rst:directive:: confluence_link
+
+    .. versionadded:: 2.1
+
+    The ``confluence_link`` directive allows a user to define a link to a
+    page that is styled with a card appearance. The directive accepts a URL.
+    How Confluence renders the context of a link card will vary based on
+    which link targets Confluence supports. For example:
+
+    .. code-block:: rst
+
+        .. confluence_link:: https://example.com
+
+    This directive supports the following options:
+
+    .. rst:directive:option:: card: card type
+        :type: block, embed
+
+        When using ``block``, a smart link card is generated for the link.
+        The card will provide a summary of the target link. When using
+        ``embed``, the contents of the link will rendered on the page.
+
+        .. code-block:: rst
+
+            .. confluence_link:: https://example.com
+                :card: block
+
+    .. rst:directive:option:: layout: layout type
+        :type: align-start, align-end, center, wrap-left, wrap-right
+
+        .. note:: This option is only applicable when using an ``embed`` card.
+
+        Specifies how an embedded card will be laid out on a page. Embedded
+        cards will default to 100% width. Therefore, to take advantage of
+        certain layout capabilities, users should also assign an appropriate
+        width as well.
+
+        .. code-block:: rst
+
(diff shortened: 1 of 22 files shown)


IceS2:
I was trying to make it work with the regular links today by passing `data-card-appearance` as inline atribute to see if maybe this PR isn't needed but I wasn't able to make it...

I'm using MystParser to parse markdown files. At some point the `data-card-appearance` attribute gets discarded, but I wasn't able to pinpoint where it is (Between Myst, Sphinx, Docutils). The Myst Parser post processor finishes with the attribute setup at the `pending_xref` tag, so I'm still digging into Sphinx to try to figure it out (=

jdknight:
@IceS2, I'm all for supporting Confluence "smart links" (v2 editor) in this extension.

The initial implementation is a good start (thanks for that). Took some time to expand on the initial implementation to provide an initial support for both web links and documentation links. Made some changes which I think may be better (but always welcome a second opinion). For "cards" that are of an inlined type, these can be created through roles; where any block-like cards will be created with directives. With the directive types, there now also provides the ability to configure embed-based cards with width hints and layout hints (which look to be available in Confluence's v2 WYSIWYG editor). I have stacked on these changes into this pull request. Feel free to try some of this out (still missing more tests for document cards, as well as documentation for users).

Some example content (which I believe can be translated to a compatible Markdown format as well; I just haven't had time to try it yet):

```
An :confluence_link:`https://www.example.com` inlined card.

.. confluence_link:: https://www.example.com
    :card: embed
    :layout: center
    :width: 80

.. confluence_link:: https://www.example.com
    :card: block


An :confluence_doc:`another-page` inlined card.

.. confluence_doc:: another-page
    :card: embed
    :layout: align-start
    :width: 40

.. confluence_doc:: another-page
    :card: block
```

IceS2:
Hey @jdknight , thanks a lot for reviewing it and adding some nice features! :partying_face: 

I think it's coming together quite nicely now. I'll try to give it a try during this week, add tests, documentation and any detail we might be missing (=

IceS2:
Hey @jdknight , I was testing out your commit and I got some extra information to share o/

## The good 

The `confluence_link` directive and role works wonderfully for the editor v2.
The `layout` and `width` attributes do nothing if you select the `block` card type but you can't edit them on the Editor neither.

## The Bad

The `confluence_doc` doesn't work for me. From what I understood it's trying to resolve it as an internal link (Like the `:doc:` role)

I'm not sure this is a good idea since you could reference other Confluence Pages that are not part of the same Sphinx Build. (You can even reference a page in another Space by passing the `ri:space-key` attribute.

I was testing it by using the following directive in markdown (MystParser way of calling directives)

```
:::{confluence_doc} 1. Record architecture decisions
:card: block
:::
```

Example on how the storage format looks from manually linking to another Space page in my confluence instance

```
<ac:link ac:card-appearance="block">
  <ri:page ri:space-key="Number2" ri:content-title="Potato" ri:version-at-save="1" />
  <ac:link-body>Potato</ac:link-body>
</ac:link>
```

You could even point to a Space home, in that case the `ri:content-title` doesn't exist.

### Interesting bit

If the page get's renamed the `ri:content-title` changes but the ID doesn't. The ID is all that we require to link to the page (Unless we're linking to an Anchor)

I believe in this scenario (Or the page moving to another Space) our documentation would be broken anyways since we're explicitely passing the parameters.
(If we were using Confluence directly the links get changed automagically)

## Thoughts

I wanted to share my findings because I'm not sure how to tackle supporting `confluence_card`. Mainly allowing the role to point to a page in another Space. All the alternatives I'm thinking about seem a bit messy.

Also I'm not sure if we should be supporting it. I like trying to cover everything but I couldn't really think of a reason to use `confluence_card` instead of `confluence_link` since on both cases we'd probably get a broken link if the page moved to another space and on the `confluence_card` scenario we'd probably get a broken link as well if the page got renamed.

What do you think?

jdknight:
> The `layout` and `width` attributes do nothing if you select the `block` card type but you can't edit them on the Editor neither.

Maybe we should have it generate a warning when using those options with `block`.

> The `confluence_doc` doesn't work for me.

That's unfortunate. Was it the role and/or the directive not working? I haven't performed much testing yet on it, but I somewhat expected the directive not not completely work since it doesn't invoke any Sphinx's resolving logic; however, I would have expected the role to work since the role type maps to Sphinx's `XRefRole` class (which should handle the work for us).

> I'm not sure this is a good idea since you could reference other Confluence Pages that are not part of the same Sphinx Build.

I don't know if I completely follow here. As you have mentioned, using `:confluence_doc:` should be akin to using `:doc:`. The reference points to some page inside the Sphinx documentation set, and the resulting Markup should reference the new/update page also published by this extension.

If the document link/reference is undefined/missing, this is a warning state and all bets are off as to what the extension will do (either a false link or no link; we make no guarantees).

Since we are references pages in our documentation set, I'm not sure how it could link to other pages.

But on that note, I do not think it would be bad if we tried to provide support for users who wish to generate document card links to other pages (even if they are not part of the publish documentation set, or even part of another space). With existing directives (e.g. `confluence_excerpt_include`) we somewhat already support this. Note that this extension makes no guarantee the link will be valid. If a documentation set provides a page title that does not exist, it is the author of the documentation that the provided links are sane. Although, I'm not sure if we want to give into this (unless this is a specific feature you are looking for).

> I believe in this scenario (Or the page moving to another Space) our documentation would be broken anyways since we're explicitely passing the parameters. (If we were using Confluence directly the links get changed automagically)

This should be fine. Confluence's storage format expects the `ri:content-title` value set to the respective page's title on publication. If the target page's name gets renamed, all other pages which point to it get updates automatically for us.

> ```
> :::{confluence_doc} 1. Record architecture decisions
> :card: block
> :::
> ```

Just to make sure, this above example would not be a valid syntax. The arguments for `confluence_doc` should be the relative document name, and not the prospect title of the target page.

Unless you are trying to provide an example on how to link to pages outside of the existing documentation set. I would imaging if we did provide the ability to link to out-of-tree pages, page identifiers (as you've mentioned) would be the right way to go. For example:

```
:confluence_doc:`~123456879`
```

or

```
:confluence_doc:`~123456879 <SPACE2>`
```

This is assuming that if we provide the raw link to Confluence, it could auto-convert it into a proper `ac:link` macro.

IceS2:
>Just to make sure, this above example would not be a valid syntax. The arguments for confluence_doc should be the relative document name, and not the prospect title of the target page.

That's fair. I thought it was supposed to link to Confluence Pages, not pages that are within the same Sphinx Project. I'll test it out as the `:doc:` role and see how it goes (=

>This should be fine. Confluence's storage format expects the ri:content-title value set to the respective page's title on publication. If the target page's name gets renamed, all other pages which point to it get updates automatically for us.

On this... that works until we republish the page without any change, right? Because then we'll force a `ri:content-title` that doesn't exist.

>Note that this extension makes no guarantee the link will be valid. 

I wouldn't expect the extension does any guarantees. It seems way out of the scope, I agree 100%! I was trying to understand how different scenarios would play out

PR #805, opened 2023-05-17
Fix missing LICENSE in sdist

maresb:
The latest [release on PyPI](https://pypi.org/project/sphinxcontrib-confluencebuilder/2.1.0/#files) is missing the `LICENSE` file. Unless I'm mistaken, this is because Flit expects the filename to be specified as the `file` argument under `license`. (The `text` argument is meant for the full license text, not the license identifier.)

I'm repackaging this for conda-forge, and this issue was caught by the infrastructure we have to make sure that all  the licenses are in order.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,7 @@ maintainers = [
 ]
 requires-python = '>=3.7'
 readme = 'README.rst'
-license = {text = 'BSD-2-Clause'}
+license = {file = 'LICENSE'}
 classifiers = [
     'Development Status :: 5 - Production/Stable',
     'Environment :: Console',


jdknight:
@maresb, thanks for bringing this up. The missing of the `LICENSE` file (and other files) were not intended (incomplete sanity checks by myself when we moved from setuptools to flit). I plan to investigate more to improve my understanding of all this (although, I won't be able to look/address this all till this evening).

I'm not sure off hand if the changes proposed are the changes we want to make here. The existing `license` value was initially defined when using setuptools (based off their [example](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html)). [Flit's documentation](https://flit.pypa.io/en/stable/pyproject_toml.html?highlight=license#new-style-metadata) indicates a different meaning for this value -- which seems odd to me. I assumed such a meta-license field is best served as a SPDX license identifier, in the way that [Poetry promotes](https://python-poetry.org/docs/pyproject/#license). Although looking at [PEP 621](https://peps.python.org/pep-0621/#license), it clearly indicates (as this PR suggests) the text of the license is desired (either explicitly or by using the file hint, which this PR uses).

The use of flit was mainly driven by Sphinx's use of this builder. Looking at their project definition, they also [use a SPDX identifier](https://github.com/sphinx-doc/sphinx/blob/master/pyproject.toml#L15) (some other projects look to be doing it the same; but... are they all wrong? or is using a SPDX ID for text more commonplace). My original thinking was to leave it as it and properly define an [include list](https://github.com/sphinx-doc/sphinx/blob/master/pyproject.toml#L111-L114) to include not only the license document, but other files that were missed in this last source distribution.

Welcome for any comments on this. In any case, the goal will be to ship out a v2.1.1 release by the end of this day.

maresb:
Hmm, I'm honestly not that experienced with Flit. And I also find the whole specification and usages of these license arguments confusing. Looking over the relevant PEP it looked like this was the appropriate fix, but I have not actually tested building it to ensure that it results in the license being included. But thanks a lot for looking into this. I'd definitely like to know what you find out.

On the conda-forge side we have an option to include the license file in the recipe, so I was nevertheless able to create a confluence builder v2.1.0 release containing the license downloaded from the repo. But I'll remove this once we get this figured out so that the provided license file takes priority and remains up-to-date.

jdknight:
Was able to finally research this a bit more. Stumbled upon [PEP 639](https://peps.python.org/pep-0639/) and its Discourse thread for what is expected soon. It appears whenever this PEP is approved, the use of `license` as a table will be deprecated. I believe we will be able to just define the following in the final version (assuming flit, or whatever builder we use, supports it):

```
license = 'BSD-2-Clause'
```

That being said, this information does not help us right now. Flit requires a table format and the existing string value could be considered incorrect. Since I'm not particularly interested in correcting the use of a key which is soon to be deprecated, I plan to submit a merge request with the `license` key removed (classifiers can still be used to reflect license information).

This does not resolve the main issue brought up with this merge request that the source distribution is missing the expected `LICENSE` document. Instead, `pyproject.toml` will most likely be updated with the following:

```
[tool.flit.sdist]
include = [
    'AUTHORS',
    'LICENSE',
    'MANIFEST.in',
]
```

This will include the `LICENSE` document, as well as other documents that should be included in the source distribution.

jdknight:
@maresb, thanks again for reporting this. Since alternative changes to `pyproject.toml` were needed, another merge request was created to address this; will close this merge request.

maresb:
Cool, thanks so much for the fix, and the detailed findings!

PR #813, opened 2023-06-08
add support for sphinxcontrib-video

AVMusorin:
Added support to process video nodes and publish assets to Confluence.

sphinxcontrib-video - https://github.com/sphinx-contrib/video


doc/features.rst
```diff
@@ -260,6 +260,7 @@ Type                              Notes
                                   time.
 `sphinxcontrib-svgbob`_           Supported
 `sphinxcontrib-youtube`_          Supported
+`sphinxcontrib-video`_            Supported
```
jdknight, line 263: (optional) Ideally, we try to these by name if possible; so this new entry should be before `sphinxcontrib-youtube`. Same with other documentation/implementation placement.

sphinxcontrib/confluencebuilder/storage/translator.py
```diff
@@ -2855,6 +2855,37 @@ def visit_vimeo(self, node):
     def visit_youtube(self, node):
         self._visit_video(node, 'https://www.youtube.com/watch?v=' + node['id'])
 
+    # ---------------------------------------------------
+    # sphinx -- extension (third party) -- sphinx-video
+    # ---------------------------------------------------
+
+    def visit_video_node(self, node):
+        autoplay = node.get("autoplay")
```
jdknight, line 2863: (optional) This extension typically uses single quotes over double quotes when possible.

sphinxcontrib/confluencebuilder/storage/translator.py
```diff
@@ -2855,6 +2855,37 @@ def visit_vimeo(self, node):
...
+        height = node.get("height")
+        width = node.get("width")
+        source_path, _ = node.get("primary_src") or (None, None)
+
+        img_key, _, _ = self.assets.add(source_path, self.docname)
```
jdknight, line 2868: (optional) Maybe replace `img_key` with `video_key` for this context.

sphinxcontrib/confluencebuilder/storage/translator.py
```diff
@@ -2855,6 +2855,37 @@ def visit_vimeo(self, node):
...
+            self.warn(f"Couldn't find video name {source_path}")
+            raise nodes.SkipNode
+
+        ri_filename = self._start_ri_attachment(node, img_key) + self._end_ri_attachment(node)
+
+        self.body.append(self._start_tag(node, 'div'))
+        self.body.append(self._start_ac_macro(node, 'multimedia'))
+        self.body.append(self._build_ac_param(node, 'name', ri_filename))
+        if width:
+            self.body.append(self._build_ac_param(node, "width", str(width)))
```
jdknight, line 2880: (optional; tangent) I wonder if we should adjust the internals of `_build_ac_param` to always `str` cast the value -- to help simplify having to use `str` in multiple locations in the translator.

sphinxcontrib/confluencebuilder/storage/translator.py
```diff
@@ -2855,6 +2855,37 @@ def visit_vimeo(self, node):
...
         self._visit_video(node, 'https://www.youtube.com/watch?v=' + node['id'])
 
+    # ---------------------------------------------------
+    # sphinx -- extension (third party) -- sphinx-video
+    # ---------------------------------------------------
+
+    def visit_video_node(self, node):
+        autoplay = node.get("autoplay")
+        height = node.get("height")
+        width = node.get("width")
```
jdknight, line 2865: The only thing I possibly see missing is handling of various height/width unit types. Under the `_visit_video` translation, the implementation utilizes `convert_length` calls to help convert Sphinx-supported units into Confluence-supported units. My guess is we would want some similar logic here as well (unless this extension on accepts a single unit type that matches Confluence's).

jdknight (commented):
This would be a nice addition; thanks for taking the time for setting up the pull request (and the documentation/test updates 🙇).

I added one comment with respect to height/width processing (if it's an issue), but other than that, it looks good. Feel free to ignore any optional comments.

jdknight:
Had time and was able to adjust the review suggestions. Thanks for the contribution to support this new third-party extension.

Issue #840, opened 2023-08-19
Issues with Sphinx 7.2.x with select custom directives/roles


jdknight:
Note that as of Sphinx 7.2.x, some performance optimizations made in Sphinx (https://github.com/sphinx-doc/sphinx/pull/11565) can cause some unexpected issues when using custom directive/roles provided by this extension. For example:

```
  File "...\sphinxcontrib\confluencebuilder\storage\translator.py", line 2732, in _visit_jira_node
    for k, v in sorted(node.params.items()):
                       ^^^^^^^^^^^
AttributeError: 'jira_issue' object has no attribute 'params'
```

There is a change in Sphinx's `_copy_except__document` where only [a subset of attributes are copied](https://github.com/sphinx-doc/sphinx/blob/76658c49a931745da45fa4d5c8d482b7c6facd4f/sphinx/util/nodes.py#L644-L648). The issue is being investigated and hopefully a fix will be made available soon.

For users blocked by this, the only recommendation at this time is to pin Sphinx to a version older than the existing v7.2.x. series. For example:

```
pip install "sphinx<7.2"
```

----

**Edit**: Users using Sphinx v7.2.x can now workaround this issue by installing the most recent development version of this extension. Fixes will be available in the next available release (v2.3) in the future.

```
pip install git+https://github.com/sphinx-contrib/confluencebuilder.git
```

jdknight:
The specific issue looks to be the use of the helper `params` attribute set on the custom nodes build in this extension. We [create the helper attribute on initialization](https://github.com/sphinx-contrib/confluencebuilder/blob/9fb5e431510b5a6782f4c198ee6559d061a66095/sphinxcontrib/confluencebuilder/nodes.py#L25). However, since the Sphinx patched copy does not copy over our helper attribute, this breaks this extension's implementation. To avoid issues, the translation will be updated to instead of using `node.params.` to use `node.attributes['params'].` instead. Also, we can drop the creation of the `.params` attribute in nodes since it is no longer possible to use longterm.

jdknight:
v2.3 is now available on PyPI -- marking as closed.

PR #841, opened 2023-08-19
explicit confluence-parameters fetch for confluence nodes

jdknight:
When accessing Confluence-extension's tracked parameters on custom nodes, always fetch this information by accessing the `attributes` attribute and using the Confluence parameter key. This is required since a helper `params` attribute may not be guaranteed to be copied by Sphinx when in the translation stage \[1\].

\[1\]: https://github.com/sphinx-doc/sphinx/blob/76658c49a931745da45fa4d5c8d482b7c6facd4f/sphinx/util/nodes.py#L633

### DIFF
--- a/sphinxcontrib/confluencebuilder/nodes.py
+++ b/sphinxcontrib/confluencebuilder/nodes.py
@@ -4,6 +4,10 @@
 from docutils import nodes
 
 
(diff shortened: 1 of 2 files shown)


jdknight:
See also #840.

PR #845, opened 2023-09-12
Add missing PARAM helper function

johannesloibl:
Refers to #841. Adds missing PARAM helper functions.


jdknight:
@johannesloibl, thanks for reporting and creating a fix for this. Any chance you can resolve the DCO check, so we can merge this in?

johannesloibl:
That should be it ;)

jdknight:
Thanks again. :bow:

PR #871, opened 2024-01-25
Update README.rst

TheDubliner:
Just updating a deprecated configuration option in the example.


jdknight:
@TheDubliner, thanks for the correction. 🙇

PR #878, opened 2024-02-20
cmd: introduce connection test

jdknight:
Provide an initial implementation of a connection-test helper, an operating mode which may help users test configuration and connection with a Confluence instance.


jdknight:
Dropping; this test branch will be worked on while attempting to debug a reported user issue. When the connection test and other capabilities help resolve this issue, a new pull request will be generated then with the updated content.

PR #913, opened 2024-03-05
Support REST v2 API Calls

jdknight:
Providing support for Confluence Cloud REST v2 APIs.

- This moves the API path prefix logic from a `Rest` class into a `Publisher` class.
- Introduces a new configuration option `confluence_api_mode` to configure which mode to use; otherwise it defaults to using `v2` for Confluence Cloud variants, or `v1` for everything else.
- The option `confluence_publish_disable_api_prefix` has been deprecated over a new option `confluence_publish_override_api_prefix`, allowing API version-specific prefix overrides.
- Updating a series of deprecated API calls with their new v2 compatible calls.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -489,6 +489,26 @@ Generic configuration
 Publishing configuration
 ------------------------
 
+.. |confluence_api_mode| replace:: ``confluence_api_mode``
+.. _confluence_api_mode:
+
+.. confval:: confluence_api_mode
+
+    Configures the API mode to use for REST requests. Certain Confluence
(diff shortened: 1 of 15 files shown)


jdknight:
A lot of changes.

Unit testing does pass, but we do not have 100% coverage. Testing perform with a validation run with Confluence Cloud appears to work. In addition, running some manual Sphinx documentation publish runs targeting Cloud and Server appear to work as well. Will attempt to perform more testing before we make a v2.5 release; and maybe others who use the main branch can report any issues from these new changes.

PR #1006, opened 2024-06-29
publisher: support confluence cloud's modernized search

jdknight:
Confluence Cloud has dropped the use of its older pagination support since 2020-07-15 \[1\] which has prevented this extension from properly processing cleanup operations if the target environment has over 249 descendants/attachments. This commit updates the search calls to use Confluence Cloud's link hints to acquire an appropriate cursor hint.

\[1\]: https://developer.atlassian.com/cloud/confluence/change-notice-moderize-search-rest-apis/


felixvd:
Thank you for this. We stumbled over this and had a local fix we were planning to submit.

jdknight:
> Thank you for this. We stumbled over this and had a local fix we were planning to submit.

@felixvd, sorry about that. Only noticed it by chance with testing in the last weeks. v2.6 of this extension has now been released (with this fix), so please feel free to try it out and report any issues.

PR #1023, opened 2024-08-05
Update development scripts/testing to include Sphinx v8.x

jdknight:
Updates the tox environment, GitHub actions and other tweaks to include Sphinx v8 series into this repository.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -19,7 +19,7 @@ release = sphinxcontrib.confluencebuilder.__version__
 supported_confluence_ver = '7.19+'
 supported_python_ver = '3.8+'
 supported_requests_ver = '2.25.0+'
-supported_sphinx_ver = '6.1+'
+supported_sphinx_ver = '7.1+'
 
 root_doc = 'contents'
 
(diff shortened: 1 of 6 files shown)


doc/conf.py
```diff
@@ -19,7 +19,7 @@
 supported_confluence_ver = '7.19+'
 supported_python_ver = '3.8+'
 supported_requests_ver = '2.25.0+'
-supported_sphinx_ver = '6.1+'
+supported_sphinx_ver = '7.1+'
```
robert-schmidtke, line 22: @jdknight Hi, the `pyproject.toml` still requires `Sphinx>=6.1`, which can break installations, as `sphinxcontrib-confluencebuilder` is upgrade to `2.7` without bumping `Sphinx` to 7.1.

See the error below.

```
Sphinx version error:
The sphinxcontrib.confluencebuilder extension used by this project needs at least Sphinx v7.1; it therefore cannot be built with this version.
```

jdknight: @robert-schmidtke, thanks for the report. Should be addressed in the v2.7.1 release.

robert-schmidtke: Thanks for resolving this so quickly!

PR #1060, opened 2024-11-06
Issue 1059: Fix downloadable abs path

### LINKED ISSUE #1059: [2.5+] Downloadable files not found anymore when using sphinx 'absolute' paths (in truth relative to 'src')
Hello,

**Issue (regression) appearing from confluencebuilder 2.5**

### Issue description

After updating confluencebuilder from version 2.4 to 2.8 (yeah, it's been a long time...) we started to get issues with `:download:` role when using an absolute path (absolute in the sense of Sphinx, i.e. the root is actually the repository root dir -env.srcdir- not the system root).

Note: It is still working fine with the HTML Builder.

Small example with a downloadable file located in a sphinx repo at: _migrated-static/files/myppt.ppt_
```
:download:`My PPT file </migrated-static/files/myppt.ppt>`
```

We get some logs like these (verbose only):
```
[confluence] failed to find path: /migrated-static/files/myppt.ppt
[confluence] process file node: /migrated-static/files/myppt.ppt
```

And a WARNING at the end:
```
WARNING: unable to find download: /migrated-static/files/myppt.ppt [confluence]
```

Could you please restore this functionnality and support these pseudo-absolute path?

### Investigation

Regression seems to come from this commit in `utils.py` file:
[Commit 336725c](https://github.com/sphinx-contrib/confluencebuilder/commit/336725c2bd869c344777205fc0cfdc018333e982#diff-d2b5691d03829fff1fa9d6002739b1e8944bf492c9e88f87d927d5f61d344eb9R216)

More specifically:
```
        # some generated nodes will prefix the path of an asset with `/`,
        # with the intent of that path being the root from the
        # configured source directory instead of an absolute path on the
        # system -- check the environment's source directory first before
        # checking the full system's path
        if normalized_path.name[0] == os.sep:
            new_path = Path(env.srcdir, normalized_path.name[1:])
```

`normalized_path.name` returns the last part of the path (name of the file) while this check was initially supposed to check the first char of the whole file path.

### Report
```
(system)
 platform: Linux-5.15.153.1-microsoft-standard-WSL2-x86_64-with-glibc2.35
   python: 3.10.12 (main, Sep 11 2024, 15:47:36) [GCC 11.4.0]
   sphinx: 8.1.3
 docutils: 0.21.2
 requests: 2.31.0
  urllib3: 2.1.0
  builder: 2.8.0

(configuration)
confluence_cleanup_purge: True
confluence_cleanup_search_mode: direct
confluence_disable_notifications: True
confluence_full_width: True
confluence_page_hierarchy: True
confluence_parent_page: (set)
confluence_publish: True
confluence_server_url: https://(removed) (cloud)
confluence_space_key: (set; upper)
confluence_timeout: 10

(confluence instance)
 connected: no
```

Paldari:
fixes #1059

This is quick fix for the regression discussed in issue 1059.


Paldari:
Ha... the test I added is obviously failing on windows since the path separator I used in the test is the Unix one... I think even on Windows a lot of people would actually use the Posix separator, so I should update the PR.

Paldari:
PR should be fine for review now.

Issue was just that a path starting with `/` or `\` is not an absolute path on windows (it must start with a drive). Kind of obvious but still took me several tries to understand without a local Windows env, haha
I could add a comment about that if you deem it necessary.

Cheers

jdknight:
Thanks for taking the time to address the issue and present a clean/signed change request, as well as reporting the issue. Sorry for the regression on this. Hopefully future changes won't break this again. Merged; and it should be available next release.

Issue #1094, opened 2025-02-20
"Unsupported Confluence API call" when using versions >2.6.0


hasier:
This issue is essentially the same as https://github.com/sphinx-contrib/confluencebuilder/issues/991, except bumping to version 2.6.0+ did not fix the issue for us (I also [wrote a comment there](https://github.com/sphinx-contrib/confluencebuilder/issues/991#issuecomment-2606984907), but figured it might be best to open a new issue).

We have experienced everything **massdosage** went through exactly as mentioned [in the original issue](https://github.com/sphinx-contrib/confluencebuilder/issues/991), except for the fact that it is not resolved for us 😅 We are currently pinned at `2.4.0`, which is fine for now, but it'd be good to be able to move to more recent versions. Please see below the debugging logs as requested previously in the original issue.

We have not tried using `v1` as [I saw in the original issue it is discouraged](https://github.com/sphinx-contrib/confluencebuilder/issues/991#issuecomment-2192858363).

<details>
  <summary>Request to fetch property</summary>

```
GET https://YOUR_SITE.atlassian.net/wiki/api/v2/pages/YOUR_PAGE_ID/properties?key=content-appearance-published

{
  "results": [
      {
        "version": {
        "number": 1,
        "message": "",
        "minorEdit": false,
        "authorId": "",
        "createdAt": "2022-07-26T06:46:57.433Z"
      },
      "value": "full-width",
      "key": "content-appearance-published",
      "id": REDACTED
    }
  ],
  "_links": {
    "base": "https://REDACTED.atlassian.net/wiki"
  }
}
```
</details>

<details>
  <summary>Report</summary>
<br/>

Running in a `ubuntu-latest` Github Action.

```
(system)
 platform: Linux-6.8.0-1017-azure-x86_64-with-glibc2.39
   python: 3.11.11 (main, Dec 4 2024, 12:58:35) [GCC 13.2.0]
   sphinx: 8.1.3
 docutils: 0.21.2
 requests: 2.32.3
  urllib3: 2.3.0
  builder: 2.9.0
(configuration)
confluence_ask_password: False
confluence_page_generation_notice: True
confluence_page_hierarchy: True
confluence_parent_page: (set)
confluence_publish: True
confluence_publish_debug: headers_and_data
confluence_server_pass: (set)
confluence_server_url: https://(removed) (cloud)
confluence_server_user: (set)
confluence_space_key: (set; upper)
(confluence instance)
 connected: yes
[...]

PR #1096, opened 2025-02-27
publisher: restore raising of an api error

jdknight:
When refactoring the retry REST call implementation \[1\], this change undesirably suppressed API errors on page update events. We should always be raising the `ConfluenceBadApiError` exception if we do not have a specific override exception to replace it with.

\[1\]: 6a1f9badd4716234271b75ac79dfe097ee6c3287

### DIFF
--- a/sphinxcontrib/confluencebuilder/publisher.py
+++ b/sphinxcontrib/confluencebuilder/publisher.py
@@ -1655,6 +1655,8 @@ class ConfluencePublisher:
                 raise ConfluenceUnreconciledPageError(
                     page_name, page['id'], self.server_url, ex) from ex
 
+            raise
+
         # post-update requests (api v2 mode)
         update_page_id = update_page['id']
 


PradnyaMhatre:
Hi @jdknight ,
Thanks for promptly responding and fixing the issue.

I checked the changes done in code, those are working as expected. (made the changes manually in publisher.py file) 

However, when I do pip install the new changes in publisher.py file are not reflecting in the package 
FYI : 
1. Tried with "pip install sphinxcontrib-confluencebuilder --upgrade"
2. Tried Uninstall-Install as well
    - python -m pip uninstall sphinxcontrib-confluencebuilder
    - python -m pip install sphinxcontrib-confluencebuilder
3. Created a new venv and did the fresh installation

But the new changes are not reflecting in the installed package

Does it usually take this long for pip installation? 
Please let me know if you have any thoughts on this.

Thanks

jdknight:
The changes are only in the development branch now. A release will most like come after any additional changes after resolving #1094. For now, I can only recommend you use the [development installation](https://sphinxcontrib-confluencebuilder.readthedocs.io/install/#development-installation) to workaround any issues you may have with the current release.

jdknight:
fyi | v2.12 is now available on PyPI which includes this change. Thanks again for reporting the issue.

PradnyaMhatre:
Thank you!

Issue #698, opened 2022-08-16
Support for comments


marshall-mcmullen:
We love confluencebuilder so much! One of the things that we are trying to find a good solution for though is that when people outside our team make in-line comments on our published confluence pages, and then we publish new docs later that night, those comments get blown away.

The comments at the bottom of the page are fine, but in-line comments get lost. And that kind of makes sense.

I'm curious if you guys have a recommended solution or workaround for this? Or if this is something that could be thought about or investigated to find a good solution or workaround.

Thank you so much for your awesome tool. We use it every day.

jdknight:
Took a look at this to see how this could be achieved. For sure there is not trivial way to address this at this time. It may be possible to support some level of inline comment retention, but it would require doing things this extension was never designed to do.

Notables are as follows:

1) Confluence at this time does not provide a way to query the contents of inline comments. So there would not be a way to (re)create comment entities if desired; however, that may not be needed for this feature. It appears when a user makes an inline comment, Confluence will inject `ac:inline-comment-marker` tags inside the source. Each of these tags includes a UUID value (inside a `ac:ref` attribute). If a user was to edit a page (or if we re-publish a page) using a comment marker with the same UUID, it will properly bind to the expected comment bubble/tree. This means we should have a way to re-inject these inline comments.

2) The one of the biggest problems with trying to re-insert comments it knowing if comments existed in the pages to be updated. To do this, page requests to Confluence (where we fetch page IDs, metadata, etc.), would now also require fetching the content of these pages (to be parsed/etc.). This could add some extra time for publishing -- which can be acceptable for users who want to maintain inline comments. This means that if we were to introduce such a capability, it would not be an automatic feature. We would most likely have a user option where one would need to explicitly configure to check and maintain comments (since we don't want to slow down other instances which do not use/care about these inline comments).

3) The hardest part is trying to find a way to diff/compare the original text with the next, and re-applying these comments. Some thought would need to be put into this -- ensuring we always use the most recent content to be published, and only inject inline comments if it is "added" and not replacing commented text. I assume there will be cases where conflicts may be detected, and the best way forward would be that any conflicting inline comments would be automatically removed.

At this time, there is no plans to implement this feature. We could possibly do something down the line to add support (next year at the earliest), but the maintenance time allocated (at least for myself) will first be focusing on providing some level of a v2 editor (experimental) support. Although if someone from the community wanted to take a stab at this, I would be happy to help review proposed merge requests to bring support for this capability.

[...]

PR #1099, opened 2025-03-16
initial/limited inlined comment check support

jdknight:
Provides a crude check to warn if a page update will result in the lost of inlined comments on a page. Users can fail a publish attempt in these situations using `--fail-on-warning` \[1\], if desired.

This commit also adds the `confluence_publish_skip_commented_pages` option, to allow skipping page updates on inlined commented pages (instead of overwriting the page).

Note that resolved comments are still considered a flagged state for inlined comment. So, a page update will warn even if all comments are resolved.

\[1\]: https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-W

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -1795,6 +1795,29 @@ Advanced publishing configuration
 
     .. versionadded:: 2.10
 
+.. confval:: confluence_publish_skip_commented_pages
+
+    .. note::
+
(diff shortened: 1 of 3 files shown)


jdknight:
For https://github.com/sphinx-contrib/confluencebuilder/issues/698.

PR #1134, opened 2025-09-03
Improved support for scoped API tokens

jdknight:
Providing support for API tokens which have scopes applied to them.
Confluence requires scoped API tokens to utilize a new API endpoint.
The changes introduced in this commit help a user to automatically
handle this endpoint change if a scoped API token is hinted.

Alternatively, users can still set the server URL to the modern
Confluence API endpoint URL. Before this change, this worked in
combination of forcing `confluence_api_mode = "v2"` to ensure v2 API 
calls are made (required since some v1 API calls do not function with
scoped APIs). Explicitly configuring a v2 API mode is no longer required
since this extension uses the scoped API-related configurations to hint
at using v2 API by default.

### DIFF
--- a/doc/advanced-publish-permissions.rst
+++ b/doc/advanced-publish-permissions.rst
@@ -1,3 +1,5 @@
+.. _publish_permissions:
+
 Publishing permissions
 ======================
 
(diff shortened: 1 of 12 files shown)


PhorstenkampFuzzy:
The docs and features look very promissing to me. The code i did not check. I can do so tomorrow if need be.
Thank you for the fast help!

jdknight:
> The docs and features look very promissing to me. The code i did not check. I can do so tomorrow if need be. Thank you for the fast help!

Thanks for looking at the documentation. This is what I was mainly interested in for a review to make sure I did not miss anything. I will go ahead and merge this in (ready for next release). That being said, if you do notice any issues, feel free to mention so.

PR #1136, opened 2025-09-26
Add option confluence_table_width

mkonig:
Default table width for editor v2 with 'widths' set is currently 760 and can not be changed.
If one has a lot of columns than 760 is just way to narrow. 
Confluence seems to be able to handle bigger width values quiet well and users might want to experiment with this value to have nicer tables.

### DIFF
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -359,6 +359,17 @@ Generic configuration
 
     .. versionadded:: 1.3
 
+.. confval:: confluence_default_table_width
+
+    Width for a table (v2 editor).
+    It has been observed that when attempting to fix specific column widths on
+    a table in the v2 editor, Confluence applies a default data width of 760 on
+    the table. By default, this value is 760.
+
+    .. code-block:: python
+
+        confluence_default_table_width = 2000
+
 .. confval:: confluence_disable_env_conf
 
     A boolean value to configure whether to ignore environment-provided
@@ -2005,19 +2016,6 @@ Advanced processing configuration
     .. code-block:: python
 
         confluence_file_suffix = '.conf'
-
-.. _confluence_table_width:
-
-.. confval:: confluence_table_width
-
-    Width for a table (v2 editor).
-    It has been observed that when attempting to fix specific column widths on
-    a table in the v2 editor, Confluence applies a default data width of 760 on
-    the table. By default, this value is 760.
-
-    .. code-block:: python
-
-        confluence_table_width = 2000
 
 .. _confluence_html_macro:
 
--- a/doc/configuration.rst
+++ b/doc/configuration.rst
@@ -2006,6 +2006,19 @@ Advanced processing configuration
 
         confluence_file_suffix = '.conf'
 
+.. _confluence_table_width:
+
+.. confval:: confluence_table_width
+
+    Width for a table (v2 editor).
+    It has been observed that when attempting to fix specific column widths on
+    a table in the v2 editor, Confluence applies a default data width of 760 on
+    the table. By default, this value is 760.
+
+    .. code-block:: python
+
+        confluence_table_width = 2000
+
 .. _confluence_html_macro:
 
 .. confval:: confluence_html_macro
--- a/sphinxcontrib/confluencebuilder/config/checks.py
+++ b/sphinxcontrib/confluencebuilder/config/checks.py
@@ -326,7 +326,8 @@ def validate_configuration(builder):
             validator.conf('confluence_default_table_width').string()
             width, unit = extract_length(config.confluence_default_table_width)
             if unit == '%':
-                raise ConfluenceDefaultTableWidthError("Percentage is not a valid unit.")
+                msg="Percentage is not a valid unit."
+                raise ConfluenceDefaultTableWidthError(msg)
         except ConfluenceConfigError as ex:
             raise ConfluenceDefaultTableWidthError(ex) from ex
 
--- a/sphinxcontrib/confluencebuilder/config/checks.py
+++ b/sphinxcontrib/confluencebuilder/config/checks.py
@@ -8,6 +8,7 @@ from sphinxcontrib.confluencebuilder.config.exceptions import ConfluenceClientCe
 from sphinxcontrib.confluencebuilder.config.exceptions import ConfluenceClientCertMissingCertConfigError
 from sphinxcontrib.confluencebuilder.config.exceptions import ConfluenceConfigError
 from sphinxcontrib.confluencebuilder.config.exceptions import ConfluenceDefaultAlignmentConfigError
+from sphinxcontrib.confluencebuilder.config.exceptions import ConfluenceDefaultTableWidthError
 from sphinxcontrib.confluencebuilder.config.exceptions import ConfluenceDomainIndicesConfigError
 from sphinxcontrib.confluencebuilder.config.exceptions import ConfluenceEditorConfigError
 from sphinxcontrib.confluencebuilder.config.exceptions import ConfluenceFooterFileConfigError
@@ -320,7 +321,10 @@ def validate_configuration(builder):
     try:
         validator.conf('confluence_default_table_width').string()
     except ConfluenceConfigError:
-        validator.conf('confluence_default_table_width').int_(positive=True)
+        try:
+            validator.conf('confluence_default_table_width').int_(positive=True)
+        except ConfluenceConfigError as ex:
+            raise ConfluenceDefaultTableWidthError(ex) from ex
 
     # ##################################################################
 
--- a/sphinxcontrib/confluencebuilder/config/checks.py
+++ b/sphinxcontrib/confluencebuilder/config/checks.py
@@ -317,8 +317,10 @@ def validate_configuration(builder):
     # ##################################################################
 
     # confluence_default_table_width
-    validator.conf('confluence_default_table_width') \
-             .int_(positive=True)
+    try:
+        validator.conf('confluence_default_table_width').string()
+    except ConfluenceConfigError:
+        validator.conf('confluence_default_table_width').int_(positive=True)
 
     # ##################################################################
 
--- a/sphinxcontrib/confluencebuilder/config/checks.py
+++ b/sphinxcontrib/confluencebuilder/config/checks.py
@@ -316,8 +316,8 @@ def validate_configuration(builder):
 
     # ##################################################################
 
-    # confluence_table_width
-    validator.conf('confluence_table_width') \
+    # confluence_default_table_width
+    validator.conf('confluence_default_table_width') \
              .int_(positive=True)
 
     # ##################################################################
--- a/sphinxcontrib/confluencebuilder/config/checks.py
+++ b/sphinxcontrib/confluencebuilder/config/checks.py
@@ -316,6 +316,12 @@ def validate_configuration(builder):
 
     # ##################################################################
 
+    # confluence_table_width
+    validator.conf('confluence_table_width') \
+             .int_(positive=True)
+
+    # ##################################################################
+
     # confluence_html_macro
     validator.conf('confluence_html_macro') \
              .string()
--- a/sphinxcontrib/confluencebuilder/config/checks.py
+++ b/sphinxcontrib/confluencebuilder/config/checks.py
@@ -44,6 +44,7 @@ from sphinxcontrib.confluencebuilder.config.validation import ConfigurationValid
 from sphinxcontrib.confluencebuilder.debug import PublishDebug
 from sphinxcontrib.confluencebuilder.std.confluence import API_MODES
 from sphinxcontrib.confluencebuilder.std.confluence import EDITORS
+from sphinxcontrib.confluencebuilder.util import extract_length
 from sphinxcontrib.confluencebuilder.util import handle_cli_file_subset
 from requests.auth import AuthBase
 import os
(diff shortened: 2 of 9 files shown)


sphinxcontrib/confluencebuilder/config/defaults.py
```diff
@@ -79,6 +79,10 @@ def apply_defaults(app):
         if conf.confluence_file_suffix.endswith('.'):
             conf.confluence_file_suffix = '.conf'
 
+    if conf.confluence_table_width:
+        if conf.confluence_table_width is None:
+            conf.confluence_table_width = 760
```
jdknight, line 84: Please restore the `CONFLUENCE_DEFAULT_V2_TABLE_WIDTH` define in its original location and replace this `720` with the define.

sphinxcontrib/confluencebuilder/storage/translator.py
```diff
@@ -3,43 +3,40 @@
...
-import re
-import sys
+from sphinxcontrib.confluencebuilder.util import (convert_length,
+                                                  extract_length, first)
 
 
 def visit_auto_context_decorator():
-    """
-    node visit decorator
+    """Node visit decorator.
```
jdknight, line 39: Please do not include/change the existing documentation style (which is mostly unrelated to the change set).

sphinxcontrib/confluencebuilder/storage/translator.py
```diff
@@ -3,43 +3,40 @@
 # Copyright 2018-2020 by the Sphinx team (sphinx-doc/sphinx#AUTHORS)
 
 from __future__ import annotations
+
+import json
+import posixpath
+import re
+import sys
```
jdknight, line 10: Please do not change the import ordering/styling used.

doc/configuration.rst
```diff
@@ -2006,6 +2006,19 @@ Advanced processing configuration
 
         confluence_file_suffix = '.conf'
 
+.. _confluence_table_width:
+
+.. confval:: confluence_table_width
```
jdknight, line 2011: Can we rename this to `confluence_default_table_width` to be somewhat similar to the existing [`confluence_default_alignment`](https://sphinxcontrib-confluencebuilder.readthedocs.io/configuration/#confval-confluence_default_alignment) configuration? We could probably also populate this under the "Generic configuration" category (beside `confluence_default_alignment`) since this is not that much of an advanced setting in the eyes of users.

sphinxcontrib/confluencebuilder/config/checks.py
```diff
@@ -316,6 +316,12 @@ def validate_configuration(builder):
 
     # ##################################################################
 
+    # confluence_table_width
+    validator.conf('confluence_table_width') \
+             .int_(positive=True)
```
jdknight, line 321: Instead of just an integer value, maybe an int or a string value should be accepted for various standard Sphinx length types used, and we use the [`extract_length`](https://github.com/sphinx-contrib/confluencebuilder/blob/v2.14/sphinxcontrib/confluencebuilder/util.py#L195)/[`convert_length`](https://github.com/sphinx-contrib/confluencebuilder/blob/v2.14/sphinxcontrib/confluencebuilder/util.py#L118) utility calls to resolve the final pixel value.

sphinxcontrib/confluencebuilder/storage/translator.py
```diff
@@ -1151,7 +1147,7 @@ def visit_table(self, node):
         # detected, apply the default table width observed when using the v2
         # editor.
         if self.v2 and 'colwidths-given' in table_classes:
-            attribs['data-table-width'] = CONFLUENCE_DEFAULT_V2_TABLE_WIDTH
+            attribs['data-table-width'] = self._v2_table_width
```
jdknight, line 1150: Note that this will only override/set the user-defined (or plugin default) only when column widths are defined on a table. If looking to be able to override defaults for all tables even non-width-hinted column tables, this would need adjustments.

jdknight: I do not believe it will cause issues under v2. We can always run the validation checks to see if it causes issues.

Maybe it would be good to introduce a `confluence_table_width` directive that operates in a similar fashion as the `highlight` directive does ([for example](https://github.com/sphinx-contrib/confluencebuilder/blob/61287ff7eaf0d7a79b3d677b2a532be9ce96a080/tests/unit-tests/datasets/code-block/code-block-highlight.rst?plain=1#L37)). So if a user wanted to override a single table's width on a document, they can do:

```
.. confluence_table_width: 500px

========= =========
Header 1  Header 2
========= =========
Row 1 Col 1 Row 1 Col 2
Row 2 Col 1 Row 2 Col 2
========= =========
```

Just food-for-thought. Does not need to be part of this pull request.

sphinxcontrib/confluencebuilder/config/defaults.py
mkonig: Done.

doc/configuration.rst
mkonig: Done.

sphinxcontrib/confluencebuilder/config/checks.py
mkonig: I did my best to make it a int/string. Hope my solution is good enough.

sphinxcontrib/confluencebuilder/storage/translator.py
```diff
@@ -1145,13 +1150,9 @@ def visit_table(self, node):
...
-        # For v2 editor, if we have given explicit widths for columns in the
-        # table (e.g. CSV table), we need to apply a data table width or the
-        # editor will ignore the column-specific widths. If widths are
-        # detected, apply the default table width observed when using the v2
-        # editor.
-        if self.v2 and 'colwidths-given' in table_classes:
-            attribs['data-table-width'] = CONFLUENCE_DEFAULT_V2_TABLE_WIDTH
+        # Apply the default table width when using the v2 editor.
+        if self.v2:
+            attribs['data-table-width'] = self._v2_default_table_width
```
jdknight, line 1155: Would it be possible to do something more like this:

```
        # Apply a user-provided default table width when using the v2 editor.
        if self.v2 and self._v2_default_table_width:
            attribs['data-table-width'] = self._v2_default_table_width
        # For v2 editor, if we have given explicit widths for columns in the
        # table (e.g. CSV table), we need to apply a data table width or the
        # editor will ignore the column-specific widths. If widths are
        # detected, apply the default table width observed when using the v2
        # editor.
        elif self.v2 and 'colwidths-given' in table_classes:
            attribs['data-table-width'] = CONFLUENCE_DEFAULT_V2_TABLE_WIDTH
```

(sorry for not being clear in the previous comments)

tests/unit-tests/test_config_checks.py
```diff
@@ -435,6 +435,25 @@ def test_config_check_header_file(self):
         self.config['confluence_header_file'] = relbase + 'sample-header.tpl'
         self._try_config()
 
+    def test_config_check_confluence_table_width(self):
+        self.config['confluence_table_width'] = 123456
```
jdknight, line 439: `confluence_default_table_width`

tests/unit-tests/test_config_checks.py
```diff
@@ -435,6 +435,25 @@ def test_config_check_header_file(self):
...
+
+        self.config['confluence_table_width'] = 0
+        with self.assertRaises(ConfluenceConfigError):
+            self._try_config()
+
+        self.config['confluence_table_width'] = -123456
+        with self.assertRaises(ConfluenceConfigError):
+            self._try_config()
+
+        self.config['confluence_table_width'] = 'MyPage'
```
jdknight, line 453: Would be nice to have a value like `250mm` and verify no exception is generated.

sphinxcontrib/confluencebuilder/storage/translator.py
```diff
@@ -125,6 +124,12 @@ def __init__(self, document, builder):
         self.colspecs = []
         self._tocdepth = self.state.toctree_depth(self.docname)
 
+        if isinstance(config.confluence_default_table_width, string):
+            table_width, twu = extract_length(config.confluence_default_table_width)
+            self._v2_default_table_width = convert_length(table_width, twu)
```
jdknight, line 129: If the resulting `data-table-width` does not take a percentage value, `pct=False` should be set in this call. If it does, please disregard.

sphinxcontrib/confluencebuilder/storage/translator.py
mkonig: This would counteract my intention of the whole change. The idea was, even if colwidths-given is true to be able to set a table width. On our company cloud hosted confluence this works fine to set both. Otherwise some of our tables are way to narrow.

If colwidths are not set then the table look weird because the columns are not streched in a sense that the full header text is in one line or in any way readable.

tests/unit-tests/test_config_checks.py
mkonig: Supposed to be done.

sphinxcontrib/confluencebuilder/storage/translator.py
mkonig: On the other hand can _v2_default_table_width be not set? Doesn't the config code set it to CONFLUENCE_DEFAULT_V2_TABLE_WIDTH if it is not set in the config?
If so it would mean the ```elif``` is never used.

sphinxcontrib/confluencebuilder/storage/translator.py
mkonig: I added a check in checks.py so that percentage is not allowed.

sphinxcontrib/confluencebuilder/config/exceptions.py
```diff
@@ -62,6 +62,17 @@ def __init__(self, msg):
...
+Accepted units include:
+
+ - px
+ - em
+ - ex
+ - mm
+ - cm
+ - in
+ - pt
+ - pc
```
jdknight, line 75: I do not believe it will be necessary to explicitly list unit sizes here. We aim to provide support for various unit sizes defined by the markup used (e.g. [rst](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#length-units)). Could write something like this:

> String values will be interpreted by units supported by the markup processed. Strings without units or integer values will be interpreted as a pixel value.

And we do not have to maintain this list as we support new unit types over time.

sphinxcontrib/confluencebuilder/storage/translator.py
jdknight: In `defaults.py`, we could change:

```
    if conf.confluence_default_table_width is None:
        conf.confluence_default_table_width = CONFLUENCE_DEFAULT_V2_TABLE_WIDTH
```

To be:

```
    if conf.confluence_default_table_width is not None:
        table_width, twu = extract_length(str(self.confluence_default_table_width))
        if table_width is not None:
            self.confluence_default_table_width = convert_length(table_width, twu, pct=False)
```

And in `storage/translator.py`, drop `self._v2_default_table_width` and directly use `confluence_default_table_width`.

sphinxcontrib/confluencebuilder/config/defaults.py
```diff
@@ -79,6 +82,12 @@ def apply_defaults(app):
         if conf.confluence_file_suffix.endswith('.'):
             conf.confluence_file_suffix = '.conf'
 
+    table_width, twu = extract_length(str(conf.confluence_default_table_width))
+    if table_width is not None:
+        conf.confluence_default_table_width = convert_length(table_width, twu, pct=False)
+    else:
+        conf.confluence_default_table_width = CONFLUENCE_DEFAULT_V2_TABLE_WIDTH
```
jdknight, line 89:
```suggestion
    if conf.confluence_default_table_width:
        table_width, twu = extract_length(str(conf.confluence_default_table_width))
        if table_width is not None:
            conf.confluence_default_table_width = convert_length(table_width, twu, pct=False)
        else:
            conf.confluence_default_table_width = CONFLUENCE_DEFAULT_V2_TABLE_WIDTH
```

sphinxcontrib/confluencebuilder/storage/translator.py
```diff
@@ -1145,13 +1144,9 @@ def visit_table(self, node):
...
-        # For v2 editor, if we have given explicit widths for columns in the
-        # table (e.g. CSV table), we need to apply a data table width or the
-        # editor will ignore the column-specific widths. If widths are
-        # detected, apply the default table width observed when using the v2
-        # editor.
-        if self.v2 and 'colwidths-given' in table_classes:
-            attribs['data-table-width'] = CONFLUENCE_DEFAULT_V2_TABLE_WIDTH
+        # Apply the default table width when using the v2 editor.
+        if self.v2:
+            attribs['data-table-width'] = config.confluence_default_table_width
```
jdknight, line 1149:
```suggestion
        # Apply the default table width when using the v2 editor.
        if self.v2 and config.confluence_default_table_width:
            attribs['data-table-width'] = config.confluence_default_table_width
        # For v2 editor, if we have given explicit widths for columns in the
        # table (e.g. CSV table), we need to apply a data table width or the
        # editor will ignore the column-specific widths. If widths are
        # detected, apply the default table width observed when using the v2
        # editor.
        elif self.v2 and 'colwidths-given' in table_classes:
            attribs['data-table-width'] = CONFLUENCE_DEFAULT_V2_TABLE_WIDTH
```

With this and the other change, you should be able to get your user-configured fixed widths from `confluence_default_table_width`. And, if `confluence_default_table_width` is not set and if `colwidths-given`, this will also avoid the need to configure a default width value for tables to allow Confluence manage this for us.

sphinxcontrib/confluencebuilder/config/defaults.py
mkonig: implemented.

mkonig:
I tried to fix the issues found by ruff. 
Especially the issue about config not being accessible. I think ruff is correct here.
I changed it to ```self.builder.config```

jdknight:
Nice; thanks for performing all the changes along the way. 🙇

Merging. Hopefully the feature helps others looking to do the same thing.

mkonig:
Thanks for your reviews, suggestions and patience.

PR #1157, opened 2025-12-11
Create py.typed: Ship types to downstream users

adamtheturtle:
I wish to override something and have to type ignore as type checkers are unaware of types defined in this project.


jdknight:
I am not that knowledgeable on the `py.typed` file. A lot of this extension's implementation is not "type ready". A majority of code that still needs to be typed/checked and there are issues (as you may have seen with the recent regressions to the mix of page ID strings and ints that is being cleaned up). My worry is having a `py.typed` class gives the impression that this code is properly typed and might be misleading.

For example, [ConfluenceAssetManager.add](https://github.com/sphinx-contrib/confluencebuilder/blob/v2.16/sphinxcontrib/confluencebuilder/assets.py#L72) does not have a type set yet:

```
def add(self, path, docname):
```

If `py.typed` is introduced and this call is not typed and used, will it cause issues?

Also, when this eventually changes to something like:

```
def add(self, path: str | bytes | os.PathLike, docname: str) -> tuple[str | None, str | None]:
```

Would the experience change?

adamtheturtle:
My understanding is that this does not imply that the code is properly or fully typed. It only means that as a downstream user I can get more errors where the type checker *does* know what is going on, and if I have strict settings I can use specific ignores for using untyped functions rather than globally disabling type checking for uses of this library.

> If py.typed is introduced and this call is not typed and used, will it cause issues?

`mypy` (with strict mode) will report `Call to untyped function "add" in typed context`.

> Also, when this eventually changes to something like: `def add(self, path: str | bytes | os.PathLike, docname: str) -> tuple[str | None, str | None]:` Would the experience change?

Yes. `mypy` would report that the `ignore` is unnecessary, and would then be able to say that a given type, for example, is incorrect.

PR #1162, opened 2026-01-08
Add configuration option for hash modifier

stibus:
Add configuration option `confluence_publish_hash_modifier` which allows to change the modifier string used when computing the hash for a page.

The option is useful e.g. when the hash values shall be retained upon changing `confluence_publish_root` or `confluence_parent_page`.

### DIFF
--- a/tests/unit-tests/test_config_postfix_formatting.py
+++ b/tests/unit-tests/test_config_postfix_formatting.py
@@ -10,8 +10,17 @@ from unittest.mock import patch
 
 
 class TestCreateDocnameUniqueHash(ConfluenceTestCase):
+    def _build_postfix_hash_modifier(self):
+        conf = self.config
+        conf.confluence_publish_postfix_hash_modifier = (
+            str(conf.project)
+            + str(conf.confluence_parent_page)
+            + str(conf.confluence_publish_root)
+        )
+
     def test_no_parent_or_root_page_or_project_configured(self):
         test_docname = 'docs/test_file.rst'
+        self._build_postfix_hash_modifier()
         hv = ConfluenceState._create_docname_unique_hash(
             docname=test_docname, config=self.config)
         self.assertEqual(hv, '75b0047d7552a5e4d91481e992f5ed339d868b3c')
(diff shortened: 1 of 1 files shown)


doc/configuration.rst
```diff
@@ -1112,32 +1136,28 @@ Publishing configuration
...
+    - :lref:`confluence_publish_hash_modifier`
 
     .. versionadded:: 1.2
     .. versionchanged:: 1.9 Support for the ``{hash}`` placeholder.
 
-.. _confluence_publish_prefix:
+.. _confluence_publish_hash_modifier:
 
-.. confval:: confluence_publish_prefix
+.. confval:: confluence_publish_hash_modifier
```
jdknight: Maybe `confluence_publish_hash_modifier` should be renamed to `confluence_publish_postfix_hash_modifier` since this is specific to the postfix hash operation (and unrelated to other page/asset hash calculations)?

doc/configuration.rst
```diff
@@ -1112,32 +1136,28 @@ Publishing configuration
...
 
-    If set, a prefix value is added to the title of all published documents. In
-    Confluence, page names need to be unique for a space. A prefix can be set to
-    either:
+    The given string will be concatenated to the *docname* of a page before
+    computing the hash.
```
jdknight: - This description does not give a clear indication to a user on why they want to manipulate the default postfix-specific hash. Can we provide more information here for users?
- Instead of `*docname*`, maybe ` ``docname`` `?

doc/configuration.rst
```diff
@@ -1112,32 +1136,28 @@ Publishing configuration
...
-       confluence_publish_prefix = 'prefix-'
+       confluence_publish_hash_modifier = (str(project)
+                                           + str(confluence_parent_page)
+                                           + str(confluence_publish_root))
 
-    By default, no prefix is used. See also:
+    See also:
 
-    - :lref:`confluence_ignore_titlefix_on_index`
     - :lref:`confluence_publish_postfix`
```
jdknight: _(optional)_ Maybe add a v3 marking in this configuration:

```
.. versionadded:: 3.0
```

jdknight (changes requested):

doc/configuration.rst
```diff
@@ -1075,6 +1075,30 @@ Publishing configuration
 
     .. versionadded:: 1.3
 
+.. _confluence_publish_prefix:
+
+.. confval:: confluence_publish_prefix
```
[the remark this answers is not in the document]
stibus: Sorry, I had not been aware of the alphabetical order. Will fix.

doc/configuration.rst
stibus: Ok.

PR #1188, opened 2026-07-12
Document label scopes required for scoped tokens

### LINKED ISSUE #1187: Label scopes missing from documentation
Using the directive below, to assign a label, fails with a scoped token with only the scopes listed in the documentation
```
.. only:: builder_confluence or builder_singleconfluence

    .. confluence_metadata::
        :labels: label-test
```

Here below the scope `write:label:confluence` is missing.  
Not sure if `read:label:confluence` is also required.  
https://github.com/sphinx-contrib/confluencebuilder/blob/b7cc6fcadc45bb940fcaa268f448797bc0844eec/doc/advanced-publish-permissions.rst?plain=1#L19-L32

oiahoon:
## Summary

- add the granular read and write label scopes to the publishing permissions reference
- keep the scoped-token setup guide synchronized with the same requirements

The Confluence content-label endpoint requires both `read:label:confluence` and `write:label:confluence` when adding labels with granular scopes.

Fixes #1187

## Verification

- `python -m sphinx -M html doc /tmp/confluencebuilder-html -E -a -W --keep-going`
- verified both generated HTML pages contain both scope identifiers
- `git diff --check`

### DIFF
--- a/doc/advanced-publish-permissions.rst
+++ b/doc/advanced-publish-permissions.rst
@@ -27,6 +27,8 @@ scopes are required:
     read:content-details:confluence
     write:attachment:confluence
     delete:attachment:confluence
+    read:label:confluence
+    write:label:confluence
     write:watcher:confluence
 
 .. references ------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


jdknight:
Thanks for closing this one up for everyone. :bow:

